PR #7, opened 2020-01-16
set DocumentSnapshot exists property

qwales1:
Ran into an issue where if I was using the `exists` property in code I was testing it would throw a run time error because `exists` is defined as a bool but since its never initialized the value returned is null. This PR sets the `exists` property to true if there is data at the path or false if not. Thank you for this library its really really helpful!


atn832:
Thank you for the pull request! I'll release a new version that includes your code tomorrow.

atn832:
https://pub.dev/packages/cloud_firestore_mocks#027

Thanks again!

Issue #6, opened 2019-12-17
Move MockCollectionReference.snapshots and where to MockQuery


atn832:
This is so that:
- `MockQueries` also fire events to the snapshots
- `where` calls can be chained: `collection.where(some condition).where(some other condition)`

PR #8, opened 2020-01-20
Implements `where` in MockQuery to enable chaining

qwales1:
I took a shot at half of issue #6 . This PR moves the where logic to its own util function and then uses that in both `MockCollectionReference.where` and `MockQuery.where` to allow chaining. Let me know what you think.


test/cloud_firestore_mocks_test.dart
```diff
@@ -109,7 +109,38 @@ void main() {
...
+      await bookmarks.add({
+        'tag': 'mostrecent',
+        'hidden': true,
+      });
+      instance
+          .collection('users')
+          .document(uid)
+          .collection('bookmarks')
+          .where('hidden', isEqualTo: false)
+          .where('tag', isEqualTo: 'mostrecent')
```
atn832: Awesome :D

atn832:
Thank you for taking this on!

This can be simplified slightly. Since `MockCollectionReference` extends `MockQuery`, and it passes all its documents to `MockQuery` on init (https://github.com/atn832/cloud_firestore_mocks/blob/master/lib/src/mock_collection_reference.dart#L28), you should be able to move the whole implementation of `where` from `MockCollectionReference` to `MockQuery` and it should work, minus some adjustments. This way you don't need to implement it twice nor move the main logic to `util`.

Something like this:

```
class MockQuery extends Mock implements Query {
  List<DocumentSnapshot> documents;
   ...
  @override
  Query where(
    dynamic field, {
    dynamic isEqualTo,
    dynamic isLessThan,
    dynamic isLessThanOrEqualTo,
    dynamic isGreaterThan,
    dynamic isGreaterThanOrEqualTo,
    dynamic arrayContains,
    List<dynamic> arrayContainsAny,
    List<dynamic> whereIn,
    bool isNull,
  }) {
    final matchingDocuments = documents
        .where((document) {
          if (isEqualTo != null) {
            return document[field] == isEqualTo;
          } else if (isGreaterThan != null) {
            Comparable fieldValue = document[field];
            if (isGreaterThan is DateTime) {
              isGreaterThan = Timestamp.fromDate(isGreaterThan);
            }
            return fieldValue.compareTo(isGreaterThan) > 0;
          } else if (isGreaterThanOrEqualTo != null) {
            Comparable fieldValue = document[field];
            if (isGreaterThanOrEqualTo is DateTime) {
              isGreaterThanOrEqualTo =
                  Timestamp.fromDate(isGreaterThanOrEqualTo);
            }
            return fieldValue.compareTo(isGreaterThanOrEqualTo) >= 0;
          } else if (isLessThan != null) {
            Comparable fieldValue = document[field];
            if (isLessThan is DateTime) {
              isLessThan = Timestamp.fromDate(isLessThan);
            }
            return fieldValue.compareTo(isLessThan) < 0;
          } else if (isLessThanOrEqualTo != null) {
            Comparable fieldValue = document[field];
            if (isLessThanOrEqualTo is DateTime) {
              isLessThanOrEqualTo = Timestamp.fromDate(isLessThanOrEqualTo);
            }
            return fieldValue.compareTo(isLessThanOrEqualTo) <= 0;
          }
          throw "Unsupported";
        })
        .toList();
    return MockQuery(matchingDocuments);
  }
}
```

It passes your test too. Would you mind tweaking your PR like so?

qwales1:
@atn832 Ah! That is way simpler. I did not notice `MockCollectionReference` extends `MockQuery`. Also I realized [this line](https://github.com/atn832/cloud_firestore_mocks/pull/8/commits/e77c45dbbe39465c48e223f7b40dbc85df4b0952#diff-22ac9481144a94a48ad8f0c184191786L51) causes problems if the value is a bool or a list so I removed it.

atn832:
Ah yes, you could have kept the line if you had written it as `final value = ...` and let the compiler make it a dynamic, or explicitly declare it as dynamic.

Thank you for the help!!

diegoveloper:
@atn832  is `whereIn` supported?

atn832:
@diegoveloper No it's not. I just filed #67 to keep track of it. You can write a PR for it if you're feeling adventurous. :)

PR #16, opened 2020-02-24
Support nested documents

### LINKED ISSUE #15: setData with keys containing dots should create nested documents
`documentReference.setData` with keys containing dots should create nested documents. Test case to confirm the expected behavior:

```
import 'package:cloud_firestore_mocks/cloud_firestore_mocks.dart';
import 'package:flutter_test/flutter_test.dart';

void main() {
    testWidgets('Nested documents created by setData with keys containing dots',
      (WidgetTester tester) async {
    final fakeFirestore = MockFirestoreInstance()..setupFieldValueFactory();

    final documentRef1 = fakeFirestore
        .collection('users')
        .document();
    final createdTime =  DateTime.now();
    await documentRef1.setData(<String, dynamic>{
      'favorites.keyA.name': 'Banana',
      'favorites.keyA.created': createdTime,
    });
    await documentRef1.setData(<String, dynamic>{
      'favorites.keyB.name': 'Apple',
      'favorites.keyB.created': createdTime,
    });
    
    final snapshot = await documentRef1.get();
    expect(snapshot['favorites'], isNotNull); // This fails.
    final favoritesMap = snapshot['favorites'] as Map<dynamic, dynamic>;
    expect(favoritesMap.keys as Iterable<String>, hasLength(2));

    final bananaMap = favoritesMap['keyA'] as Map<dynamic, dynamic>;
    expect(bananaMap['name'] as String, 'Banana');
    expect(bananaMap['created'] as DateTime, createdTime);
  });
}

```

Screenshot from real Firebase (from a different example):

Screen Shot 2020-02-23 at 11 48 49 PM

I might contribute a PR to this issue.

suztomo:
Fixes #15 .

@atn832 Would you review this?  This still needs to handle more edge cases such as deleting the last field of a object, but I'd like to confirm the overall direction with you.

### DIFF
--- a/lib/src/mock_document_reference.dart
+++ b/lib/src/mock_document_reference.dart
@@ -30,22 +30,49 @@ class MockDocumentReference extends Mock implements DocumentReference {
   @override
   Future<void> updateData(Map<String, dynamic> data) {
     data.forEach((key, value) {
+      final documentToUpdate = _findNestedDocumentToUpdate(key);
+      if (documentToUpdate != root) {
+        key = key.split('.').last;
+      }
       if (value is FieldValue) {
         final fieldValuePlatform = FieldValuePlatform.getDelegate(value) as MockFieldValuePlatform;
         switch (fieldValuePlatform.value) {
           case MockFieldValue.delete:
-            root.remove(key);
+            documentToUpdate.remove(key);
             break;
           default:
             throw Exception('Not implemented');
         }
       } else if (value is DateTime) {
-        root[key] = Timestamp.fromDate(value);
+        documentToUpdate[key] = Timestamp.fromDate(value);
       } else {
-        root[key] = value;
+        documentToUpdate[key] = value;
       }
     });
     return Future.value(null);
+  }
+
+  Map<String, dynamic> _findNestedDocumentToUpdate(String keyWithPeriods) {
+    final keyElements = keyWithPeriods.split('.');
+    if (keyElements.length == 1) {
+      return root;
+    }
+
+    Map<String, dynamic> document = root;
+    
+    // For N elements, iterate until N-1 element.
+    // For example, key: "foo.bar.baz", this method return the document pointed by
+    // 'foo.bar'. The document will be updated by the caller to have key 'baz'
+    final keysToIterate = keyElements.sublist(0, keyElements.length -1 );
+    for (String key in keysToIterate) {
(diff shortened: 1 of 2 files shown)


atn832:
This looks good!

Btw, if it were recursive, would it be more readable and help handle your deletion edge case? Something like:

```dart
@override
Future<void> updateData(Map<String, dynamic> data) {
  data.forEach((key, value) {
    _updateData(root, key, value);
  });
  return Future.value(null);
}

_updateData(Map<String, dynamic> data, String key, dynamic value) {
  // Base case.
  if (key has no dot) {
    // handle FieldValue case or
    data[key] = value;
    return;
  }

  // Handle "composite" keys.
  final tokens = key.split('.');
  final key = tokens.removeAt(0);
  final subkey = tokens.join('.');
  if (!data.containsKey(key)) {
    // create the submap
  }
  final submap = data[key];
  _updateData(submap, subkey, value);

  // your empty edge case
  if (submap.isEmpty) {
    data.remove(key);
  }
}
```

suztomo:
Regarding the deletion edge case, it was my wrong assumption. 

Firestore allows empty documents. So no need to delete a document upon the last field removal. Therefore no need to make recursive calls.

Screen Shot 2020-02-25 at 9 45 32 AM
Screen Shot 2020-02-25 at 9 45 42 AM

lib/src/mock_document_reference.dart
```diff
@@ -30,30 +30,62 @@ class MockDocumentReference extends Mock implements DocumentReference {
...
+
+    Map<String, dynamic> document = root;
+
+    // For N elements, iterate until N-1 element.
+    // For example, key: "foo.bar.baz", this method return the document pointed by
+    // 'foo.bar'. The document will be updated by the caller on 'baz' field
+    final keysToIterate =
+        compositeKeyElements.sublist(0, compositeKeyElements.length - 1);
+    for (String keyElement in keysToIterate) {
+      if (!document.containsKey(keyElement) || !(document[keyElement] is Map)) {
```
suztomo: When a field is not a map, then Firestore makes it as a map.

Screen Shot 2020-02-25 at 10 29 27 AM

atn832: Interesting behavior. I guess the opposite is true too? If it's a map and they set a non-map value instead, it becomes just that non-map value.

PR #28, opened 2020-03-01
Batch to support updateData and delete operation

### LINKED ISSUE #25: Document and write tests for Firestore.batch
(I haven’t tested but I assume this is not implemented. I’m going to use batch in my Flutter app)

Similar to a transaction but a batch holds only write operations.

Example usage of batch
https://github.com/FirebaseExtended/flutterfire/blob/master/packages/cloud_firestore/cloud_firestore/example/lib/main.dart#L110

suztomo:
Fixes #25


atn832:
Thanks for implementing the 2 other missing methods!

PR #29, opened 2020-03-01
addData to use automatic documentID generation

### LINKED ISSUE #23: Idea: the test not to rely on the predictable generation key logic “z, zz, zzz...”
Follow up of https://github.com/atn832/cloud_firestore_mocks/pull/22.

DocumentSnapshotMatcher may optionally ignore the documentID field when the calling test does not need to assert on the ID.

Variables “expectedDumpAfterXXXX” can move to local variables and take `${addedDocument.documentID}` in the String.

suztomo:
Fixes #23

- addData to use automatic documentID generation
- New `DocumentSnapshotMatcher.onData` to create a matcher that does not verify documentID.

### DIFF
--- a/test/cloud_firestore_mocks_test.dart
+++ b/test/cloud_firestore_mocks_test.dart
@@ -13,28 +13,6 @@ const expectedDumpAfterSetData = """{
   }
 }""";
 
-const expectedDumpAfterAddData = """{
-  "messages": {
-    "z": {
-      "content": "hello!",
-      "uid": "abc"
-    }
-  }
-}""";
-
-const expectedDumpAfterSuccessiveAddData = """{
-  "messages": {
-    "z": {
-      "content": "hello!",
-      "uid": "abc"
-    },
-    "zz": {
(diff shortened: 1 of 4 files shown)


test/cloud_firestore_mocks_test.dart
```diff
@@ -348,7 +345,7 @@ void main() {
 
     final snapshot2 = await firestore.collection('users').document().get();
     expect(snapshot2, isNotNull);
-    expect(snapshot2.documentID.length, 20);
+    expect(snapshot2.documentID.length, greaterThanOrEqualTo(20));
```
suztomo: The auto-generated key length might not exact 20. In real Firestore, I got `/users/t1gVCkBrgMU9LjFRnrtu1FFjlir1/persons/DppyyRuUwQL7W6Qq7t8O` . The first key has 28 characters and the second has 20.

atn832:
Much cleaner than my old loop. Thank you!

Issue #21, opened 2020-02-26
documentSnapshot.data should be null before the document is saved


suztomo:
Test case:

```
    final snapshot1 =
        await firestore.collection('users').document(nonExistentId).get();
    expect(snapshot1, isNotNull);
    expect(snapshot1.documentID, nonExistentId);
    // data field should be null before the document is saved
    expect(snapshot1.data, isNull);
```
As of now, this test fails because `snapshot1.data` is an empty Map. Real Firestore has null there.

<img width="580" alt="Screen Shot 2020-02-25 at 7 41 05 PM" src="https://user-images.githubusercontent.com/28604/75303664-5ca08980-580f-11ea-81a1-45ce069cdfce.png">

PR #32, opened 2020-03-02
MockFirestoreInstance to remember saved documents

### LINKED ISSUE #20: collection('...').document() should not save a document
With current implementation, the following test code fails:
```
    final reference1 = firestore.collection('users').document();
    final reference2 = firestore.collection('users').document();

    await reference1.setData({
      'someField': 'someValue',
    });

    QuerySnapshot querySnapshot = await firestore.collection('users').getDocuments();
    // Only reference1 is saved
    expect(querySnapshot.documents, hasLength(1));
    expect(querySnapshot.documents.first['someField'], 'someValue');
```

It fails because as of now cloud_firestore_mocks does not distinguish saved and unsaved documents.

I suspect code below should be updated to accommodate this.
```
  @override
  DocumentReference document([String path]) {
    if (path == null) {
      path = _generateAutoId();
    }

    return MockDocumentReference(path, getSubpath(root, path), root,
        getSubpath(snapshotStreamControllerRoot, path));
  }
```

`getSubpath(root, path)` creates a document to the root.

suztomo:
Fixes #20 and #21 .

### DIFF
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -13,7 +13,7 @@ import 'util.dart';
 class MockFirestoreInstance extends Mock implements Firestore {
   Map<String, dynamic> _root = Map();
   Map<String, dynamic> _snapshotStreamControllerRoot = Map();
-
+  final Set _savedDocumentPaths = Set<String>();
   MockFirestoreInstance() {
     _setupFieldValueFactory();
   }
@@ -39,6 +39,14 @@ class MockFirestoreInstance extends Mock implements Firestore {
     JsonEncoder encoder = JsonEncoder.withIndent('  ', myEncode);
     final jsonText = encoder.convert(_root);
     return jsonText;
+  }
+
+  void saveDocument(String path) {
+    _savedDocumentPaths.add(path);
+  }
+
+  bool hasSavedDocument(String path) {
+    return _savedDocumentPaths.contains(path);
   }
 
   _setupFieldValueFactory() {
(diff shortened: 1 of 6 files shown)


suztomo:
Will resolve conflict.

test/cloud_firestore_mocks_test.dart
```diff
@@ -140,6 +140,84 @@ void main() {
...
+    expect(documentReference.path, 'users/aaa/friends/bbb/friends-friends/ccc');
+  });
+
+  test('Creating document reference should not save the document', () async {
+    final instance = MockFirestoreInstance();
+    await instance.collection('users').add(<String, dynamic>{'name': 'Foo'});
+    final documentReference = instance.collection('users').document(uid);
+
+    var querySnapshot = await instance.collection('users').getDocuments();
+    expect(querySnapshot.documents, hasLength(1));
```
suztomo: This is the very motivation of this PR.

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -40,6 +44,18 @@ class MockFirestoreInstance extends Mock implements Firestore {
     return jsonText;
   }
 
+  void saveDocument(String path) {
```
atn832: Consider moving all the saveDocument functionality to a class that is not exposed, so that it doesn't show up in the documentation.

atn832: I've filed #36 to keep track of it.

atn832:
Published this, along with your other PRs to https://pub.dev/packages/cloud_firestore_mocks#041. Thank you as always!

suztomo:
@atn832 Thank you.

PR #38, opened 2020-03-09
MockQuery to defer query execution until getDocuments

### LINKED ISSUE #31: MockQuery should defer query execution
MockQuery should defer query execution. In this example code, `subcollection` is updated twice. In real Firebase, querying the same `subcollection` should work.

```
  test('Saving documents in subcollection', () async {
    await instance
        .collection('users')
        .document(uid)
        .collection('friends')
        .add(<String, dynamic>{'name': 'Foo'});

    final documentReference = instance
        .collection('users')
        .document(uid)
        .collection('friends')
        .document();

    var subcollection =
        instance.collection('users').document(uid).collection('friends');
    var querySnapshot = await subcollection.getDocuments();
    expect(querySnapshot.documents, hasLength(1));

    // Only after setData, the document is available for getDocuments
    await documentReference.setData({'name': 'Bar'});

    // This line below should be unnecessary.
    subcollection =
        instance.collection('users').document(uid).collection('friends');
    querySnapshot = await subcollection.getDocuments();
    expect(querySnapshot.documents, hasLength(2));
```

suztomo:
Fixes #31 

MockQuery should not hold the query results; otherwise it may return stale data.

In my use case, my class hold a collection reference (a final field) during the whole lifetime of my app. The class adds and retrieves documents from the Firestore collection through the the same reference.

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -1,45 +1,63 @@
 import 'dart:math';
 
 import 'package:cloud_firestore/cloud_firestore.dart';
-import 'package:cloud_firestore_mocks/cloud_firestore_mocks.dart';
 import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_interface.dart';
 import 'package:mockito/mockito.dart';
 
 import 'mock_snapshot.dart';
 
-class MockQuery extends Mock implements Query {
-  List<DocumentSnapshot> documents;
-  final MockFirestoreInstance _firestore;
+typedef List<DocumentSnapshot> _QueryOperation(List<DocumentSnapshot> input);
 
-  MockQuery(this._firestore, this.documents);
+class MockQuery extends Mock implements Query {
+  /// Previous query in a Firestore query chain. Null if this instance is a
+  /// collection reference. A query chain always starts with a collection
+  /// reference, which does not have a previous query.
+  final Query _parentQuery;
+
+  /// Operation to perform in this query, such as "where", "limit", and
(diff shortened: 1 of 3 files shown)


test/mock_query_test.dart
```diff
@@ -196,4 +196,57 @@ void main() {
...
+  test('Collection reference should not hold query result', () async {
+    final instance = MockFirestoreInstance();
+
+    final collectionReference = instance.collection('users');
+    await collectionReference.add({
+      'username': 'Bob',
+    });
+    final snapshot = await collectionReference.getDocuments();
+    expect(snapshot.documents, hasLength(1));
+  });
```
suztomo: This test case is the very motivation of this PR. It was failing before this PR.

lib/src/mock_query.dart
```diff
@@ -1,45 +1,63 @@
...
 
-  MockQuery(this._firestore, this.documents);
+  MockQuery([this._parentQuery, this._operation]);
 
   // ignore: unused_field
   final QueryPlatform _delegate = null;
 
   @override
-  Future<QuerySnapshot> getDocuments({Source source = Source.serverAndCache}) {
-    final savedDocuments = documents.where((snapshot) => _firestore.hasSavedDocument(snapshot.reference.path)).toList();
```
suztomo: Now Query is not responsible to check whether a document is saved or not. It is collection references' responsibility to check that, and a collection reference is always at the beginning of a query chain.

PR #45, opened 2020-03-24
Updating a document should not affect previously-retrieved snapshots

### LINKED ISSUE #44: Updating a document should not affect previous snapshots
Test case:

```
  test('Snapshot should remain after updating data', () async {
    final firestore = MockFirestoreInstance();
    // These documents are not saved
    final reference = firestore.collection('users').document('foo');
    await reference.setData(
      <String, dynamic>{'name': 'old'}
    );
    final snapshot1 = await reference.get();

    await reference.setData(
        <String, dynamic>{'name': 'new'}
    );

    expect(snapshot1.data['name'], 'old');
  });
```

I guess I'll need to implement deep copy of Firestore documents.

suztomo:
Fixes #44 .

### DIFF
--- a/example/test_driver/cloud_firestore_behaviors.dart
+++ b/example/test_driver/cloud_firestore_behaviors.dart
@@ -183,5 +183,31 @@ void main() {
       expect(nestedData2['message'], 'updated value');
       expect(nestedData2['unaffected_field'], 'old value2');
     });
+
+    ftest('Snapshot should not be affected by updates', (firestore) async {
+      final CollectionReference messages = firestore.collection('messages');
+
+      final DocumentReference doc = messages.document();
+      // updateData requires an existing document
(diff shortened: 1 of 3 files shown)


example/test_driver/cloud_firestore_behaviors.dart
```diff
@@ -183,5 +183,31 @@ void main() {
...
+
+      await doc.setData({'foo': 'new'});
+      await doc.updateData(<String, dynamic>{
+        'nested.data.message': 'new nested data',
+      });
+
+      await doc.delete();
+
+      // At the time the snapshot was created, the value was 'old'
+      expect(snapshot.data['foo'], 'old');
```
suztomo: This assertion was failing for cloud_firestore_mocks before this PR.

atn832:
:shipit:

PR #54, opened 2020-03-28
FieldValue integration tests

suztomo:
This drive test has two invocation types with different environment variable `FIRESTORE_IMPLEMENTATION`:

```
~/Documents/cloud_firestore_mocks $ FIRESTORE_IMPLEMENTATION=cloud_firestore\
 flutter drive --target=test_driver/field_value_behaviors.dart &&  \
FIRESTORE_IMPLEMENTATION=cloud_firestore_mocks \
flutter drive --target=test_driver/field_value_behaviors.dart
...
flutter: Skipping cloud_firestore_mocks
flutter: 00:04 +13: Firestore behavior on FieldValue: FieldValue.arrayRemove (Cloud Firestore)
flutter: 00:05 +14: Firestore behavior on FieldValue: FieldValue.arrayRemove (Firestore Emulator)
flutter: 00:05 +15: (tearDownAll)
flutter: 00:05 +16: All tests passed!
...
flutter: 00:00 +14: Firestore behavior on FieldValue: FieldValue.arrayRemove (Firestore Emulator)
flutter: Skipping Firestore Emulator
flutter: 00:00 +15: (tearDownAll)
flutter: 00:00 +16: All tests passed!
Stopping application instance.
```

PR #55, opened 2020-03-29
Copy Documents upon Save

suztomo:
This should be done after #54 is merged.

Before this PR, `setData` did not copy Map and Arrays when saving data. Therefore, modification to the data passed as the argument of `setData`, `add`, and `updateData` resulted in affecting the documents held by cloud_firestore_mocks.


suztomo:
This should be done after #54 is merged. Closing this for now.

suztomo:
This is superseded by #56

PR #62, opened 2020-04-05
getSubpath to handle path that contains slashes

### LINKED ISSUE #61: DocumentReference through FirestoreInstance.document() does not save document
This test fails at 0.4.2:

```
  test('Saving documents through FirestoreInstance.document()',
      () async {
    // Creates 1st document in "users/abc/friends/<documentId>"
    await instance.document('users/$uid/friends/xyz').setData({
      'name': 'Foo',
      'nested': {
        'k1': 'v1',
      }
    });

    final documentReference = instance
        .collection('users')
        .document(uid)
        .collection('friends')
        .document('xyz');

    final snapshot = await documentReference.get();
    expect(snapshot.data['name'], 'Foo');
    final nested = snapshot.data['nested'] as Map<String, dynamic>;
    expect(nested['k1'], 'v1');
  });
```

suztomo:
Fixes #61

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -1,10 +1,19 @@
 import 'package:cloud_firestore/cloud_firestore.dart';
 
 dynamic getSubpath(Map<String, dynamic> root, String path) {
-  if (root[path] == null) {
-    root[path] = Map<String, dynamic>();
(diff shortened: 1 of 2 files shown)


atn832:
Thank you!

PR #69, opened 2020-04-09
add support for arrayContains in where

qwales1:
Adds support for `arrayContains`. This assumes that the correct behavior for using arrayContains on a field that is not of an array type is to not match any documents. This appears to be how it works in the JS SDK at least but I could not find a reference that explicitly states that.

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -122,6 +122,12 @@ class MockQuery extends Mock implements Query {
         isLessThanOrEqualTo = Timestamp.fromDate(isLessThanOrEqualTo);
       }
       return fieldValue.compareTo(isLessThanOrEqualTo) <= 0;
+    } else if (arrayContains != null) {
+      if (value is Iterable) {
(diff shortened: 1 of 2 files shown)


atn832:
Very nice. That's exactly what the documentation for the Android library states ([link](https://firebase.google.com/docs/reference/android/com/google/firebase/firestore/Query#whereArrayContains(com.google.firebase.firestore.FieldPath,%20java.lang.Object))).

I'll submit your PR now. When you get a chance, could you paste the same unit tests into https://github.com/atn832/cloud_firestore_mocks/blob/master/test_driver/cloud_firestore_behaviors.dart? This ensures that cloud_firestore_mocks behaves the same as the real Firestore and Firestore emulator. I've filed #71 to keep track of it.

Thanks again for your pull request, @qwales1!

atn832:
Btw, I just published your PR. It is live at https://pub.dev/packages/cloud_firestore_mocks#043. Thanks!

PR #75, opened 2020-04-19
Add support for arrayContainsAny query

### LINKED ISSUE #74: Missing support for query.arrayContainsAny
[MockQuery](https://github.com/atn832/cloud_firestore_mocks/blob/master/lib/src/mock_query.dart) presently supports an impressive list of actual Firestore queries. The one that is sorely missing is `arrayContainsAny`, which is like doing multiple `arrayContains` queries simultaneously.

I believe providing an implementation for this will not be too complicated or time consuming. Please consider adding support for it.

anuragbhd:
Fixes #74

This change will allow to use arrayContainsAny operator with `query.where`

See official Firestore documentation for more details:
https://firebase.google.com/docs/firestore/query-data/queries#in_and_array-contains-any


atn832:
Thank you for writing this PR @anuragbhd! Would you mind writing unit tests as well to verify that the code works as expected? You can take a look at how @qwales1 did it for arrayContains in #69.

anuragbhd:
Thanks for the pointer, @atn832. I've written tests based on existing ones for `arrayContains`, and found them to be all green :)

test/mock_query_test.dart
```diff
@@ -188,6 +188,39 @@ void main() {
...
+      'name': 'Post #3',
+      'tags': ['mostrecent'],
+    });
+    await instance.collection('posts').add({
+      'name': 'Post #4',
+      'tags': ['mostrecent', 'interesting'],
+    });
+    instance
+        .collection('posts')
+        .where('tags', arrayContainsAny: ['interesting', 'mostrecent'])
```
atn832, line 211: According to https://firebase.google.com/docs/firestore/query-data/queries#in_and_array-contains-any, array-contains-any should return documents where the value is an array that contains one or more values of the comparison values. So shouldn't it return all 4 documents?

lib/src/mock_query.dart
```diff
@@ -128,6 +128,12 @@ class MockQuery extends Mock implements Query {
       } else {
         return false;
       }
+    } else if (arrayContainsAny != null) {
+      if (value is Iterable) {
+        return Set.from(value).containsAll(arrayContainsAny);
```
atn832, line 133: According to the specs, we should be checking that value contains at least one element in arrayContainsAny, not all of them.

test/mock_query_test.dart
anuragbhd: Thanks for pointing that out. My bad. I understood Firestore's `array-contains-any` incorrectly. Too bad it doesn't support something like `array-contains-all`, which is what I was looking for.

lib/src/mock_query.dart
anuragbhd: Yep, my latest commit addresses this. Tests have been updated accordingly.

anuragbhd:
Added one more commit to enforce the condition that `arrayContainsAny` must not have more than 10 comparison values. Update unit tests accordingly.

This one is optional. Ignore it if you fee like.

lib/src/mock_query.dart
```diff
@@ -128,6 +128,23 @@ class MockQuery extends Mock implements Query {
...
         return false;
       }
+    } else if (arrayContainsAny != null) {
+      if (arrayContainsAny.length > 10) {
+        throw ArgumentError(
+          'arrayContainsAny cannot contain more than 10 comparison values',
+        );
+      }
+      if (value is Iterable) {
+        var valueSet = Set.from(value);
```
atn832, line 138: minor detail: this could have been `final`.

atn832:
Awesome tests. And thanks again for writing the PR. I'll publish a new version of the library that includes it, this weekend.

atn832:
Your code is live at https://pub.dev/packages/cloud_firestore_mocks#043. Thank you @anuragbhd !

PR #72, opened 2020-04-11
Adds driver test for arrayContains

qwales1:
Fixes #71 

I am a little unclear on how to run this test on the android emulator to make sure its working. The [readme](https://github.com/atn832/cloud_firestore_mocks/tree/master/test_driver#run-driver-test) says to run in the example folder but the `cloud_firestore_behaviors.dart` file doesn't exist there.

PR #80, opened 2020-04-24
Make MockDocumentSnapshot data immutable

### LINKED ISSUE #79: Map returned from `MockDocumentSnapshot.data` should be immutable
The object returned from the actual Firestore implementation of `DocumentSnapshot.data` getter is a `Map<String, dynamic>` that apparently does not respond to value changes.

For example, something like this presently works with cloud_firestore_mocks but not with the real [cloud_firestore](https://pub.dev/packages/cloud_firestore):

```dart
final snapshot = await store.collection('products').document(productId).get();
// snapshot.data is a Map like { 'id': null, 'name': 'iPhone SE', 'price': 399.0 }
snapshot.data['id'] = snapshot.documentID;
return snapshot.data; // 'id' in returned Map is still null (with real Firestore)
```

A way around with real Firestore is:

```dart
final snapshot = await store.collection('products').document(productId).get();
final data = snapshot.data;
data['id'] = snapshot.documentID;
return data; // 'id' in returned Map is now correctly set
```

This is weird as `snapshot.data` and the new `data` variables both essentially point to the same Map. But it works for some reason.

This is a small quirk that `MockDocumentSnapshot.data` should also handle. Because of the current implementation, my unit tests that expect "id" to be set correctly passed with cloud_firestore_mocks but my logic with cloud_firestore failed.

anuragbhd:
Fixes #79 

This is likely not the right way to fix the referred issue. I am open to suggestions :)

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -2,6 +2,8 @@ import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:cloud_firestore_mocks/src/mock_document_reference.dart';
 import 'package:mockito/mockito.dart';
 
+import 'util.dart';
+
 class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   final String _documentId;
   final Map<String, dynamic> _document;
@@ -22,7 +24,7 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   @override
   Map<String, dynamic> get data {
     if (_exists) {
-      return Map<String, dynamic>.unmodifiable(_document);
+      return deepCopy(_document);
     } else {
       return null;
     }
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -22,7 +22,7 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   @override
   Map<String, dynamic> get data {
     if (_exists) {
-      return _document;
+      return Map<String, dynamic>.unmodifiable(_document);
     } else {
       return null;
     }
--- a/test/cloud_firestore_mocks_test.dart
+++ b/test/cloud_firestore_mocks_test.dart
@@ -842,15 +842,18 @@ void main() {
     expect(erroneousTransactionUsage, throwsA(isA<PlatformException>()));
   });
 
-  test('Document snapshot data is unmodifiable', () async {
+  test('Document snapshot data returns a new instance', () async {
     final instance = MockFirestoreInstance();
     await instance.collection('users').document(uid).setData({
-      'name': 'Bob',
+      'name': 'Eve',
+      'friends': ['Alice', 'Bob'],
     });
 
-    final bob = await instance.collection('users').document(uid).get();
-    bob.data['name'] = 'John';
+    final eve = await instance.collection('users').document(uid).get();
+    eve.data['name'] = 'John';
+    eve.data['friends'][0] = 'Superman';
 
-    expect(bob.data['name'], isNot('John'));
+    expect(eve.data['name'], isNot('Superman')); // nothing changed
+    expect(eve.data['friends'], equals(['Alice', 'Bob'])); // nothing changed
   });
 }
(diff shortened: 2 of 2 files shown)


atn832:
It looks like the correct way to fix the issue.

Can you also add the same test to the integration tests at https://github.com/atn832/cloud_firestore_mocks/blob/38c30d477399648551a589ffe2fc6fd239b21a77/test_driver/cloud_firestore_behaviors.dart#L376? See #72 for an example of how an integration test was added (it's really just a copy/paste). Once this is done, we can run this test against both the real Firestore and your implementation and verify that they both behave the same way.

atn832:
Since running integration tests are a bit tricky, I can run them for you on my machine and check that it all works.

anuragbhd:
> It looks like the correct way to fix the issue.
> 

The problem with my current implementation is that doing something like this throws an exception:

```dart
final bob = await instance.collection('users').document(uid).get();
bob.data['name'] = 'John'; // throws "Unsupported operation: Cannot modify unmodifiable map"
```

Instead, changing a key's value in `data` map should just have no effect, without an error. That is how Firestore behaves.

Furthermore, the following should work:

```dart
final bob = await instance.collection('users').document(uid).get();
final data = bob.data;
data['name'] = 'John'; // should actually update the value
```

Again, this doesn't work with an UnmodifiableMap. Perhaps this can be solved with another data structure. Which one, I do not know.

anuragbhd:
> Can you also add the same test to the integration tests at
> 
> https://github.com/atn832/cloud_firestore_mocks/blob/38c30d477399648551a589ffe2fc6fd239b21a77/test_driver/cloud_firestore_behaviors.dart#L376
> 
> ? See #72 for an example of how an integration test was added (it's really just a copy/paste). Once this is done, we can run this test against both the real Firestore and your implementation and verify that they both behave the same way.

Sure, I'll update the integration tests once we've finalized a proper implementation.

atn832:
This behavior of Firestore is quite strange, and I can't explain why it won't update the map in your first example but will in the second one. A few options come to mind:
1) this behavior can be really misleading to Firestore users. We can consider filing a bug to them and keeping our implementation as is.
2) check the runtime type of what's returned by Firestore, and use the same class if possible. The problem is that this behavior is probably not documented. Not sure we'd want to replicate it.
3) I'd prefer staying away from this last option: implement the same strange behavior.

I'll dig more this weekend.

anuragbhd:
> 1. this behavior can be really misleading to Firestore users. We can consider filing a bug to them and keeping our implementation as is.

Yep. I'll try and see if I can raise an issue soon over at [FlutterFire](https://github.com/FirebaseExtended/flutterfire).

> I'll dig more this weekend.

That'd be great, thanks.

atn832:
I figured it out. Firestore is working as intended. In the Firestore implementation, the `data` getter returns a new copy of the data every time it's called. That's why if you check `snapshot.data` after modifying `snapshot.data['id']`, you won't see your modification. However cloud_firestore_mocks returns the same instance every time.

I've also checked that Firestore returns a deep copy by testing on a list value.

So the fix is to return a deep copy of `_document`. Luckily, @suztomo wrote a method you can reuse directly at https://github.com/atn832/cloud_firestore_mocks/blob/38c30d477399648551a589ffe2fc6fd239b21a77/lib/src/util.dart#L31.

You can use this test case to make sure that we're returning a deep copy, and not a shallow copy:
```
  test('Document snapshot data returns a new instance', (firestore) async {
    await firestore.collection('users').document(uid).setData({
      'friends': ['Alice', 'Bob'],
    });
    final bob = await firestore.collection('users').document(uid).get();
    bob.data['friends'][0] = 'John';
    expect(bob.data['friends'], equals(['Alice', 'Bob']));
  });
```

Would you mind modifying your CL? We're nearly there!

anuragbhd:
@atn832 Great find! Thanks for checking on my behalf. I've pushed a new commit to address this.

atn832:
Looks good, thank you!!

In order not to delay this PR any further, and since I've manually confirmed Firestore's behavior last week, I'll merge it now. I'll add the same test to the integration tests suite later.

PR #82, opened 2020-04-24
Add nested object support in MockQuery

### LINKED ISSUE #81: Nested object support in MockQuery
```
  await fs.collection("tests").add({"testSubObject": {"key": "abc"}});

var result = await fs.collection("tests").where("testSubObject.key", isEqualTo: "abc").getDocuments();

    print(result.documents.length);
```

Right now this returns 0 documents, whilst in FireStore it returns 1 document. 

This has lead to some of our code becoming impossible to test with cloud_firestore_mocks, though I am not sure how many people actually use this feature...

C-Alexander:
Fixes #81

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -90,7 +90,7 @@ class MockQuery extends Mock implements Query {
       List<dynamic> whereIn,
       bool isNull}) {
     _QueryOperation operation = (documents) => documents
-        .where((document) => _valueMatchesQuery(document[field],
+        .where((document) => _valueMatchesQuery(_getValueFromDocument(document, field),
             isEqualTo: isEqualTo,
             isLessThan: isLessThan,
             isLessThanOrEqualTo: isLessThanOrEqualTo,
@@ -165,5 +165,20 @@ class MockQuery extends Mock implements Query {
       }
     }
     throw "Unsupported";
+  }
+
+  dynamic _getValueFromDocument(dynamic document, dynamic field) {
+    if (field.runtimeType != String) {
+      return document[field];
+    }
+
+    var subFields = (field as String).split(".");
+
+    var value = document;
+    for (var field in subFields) {
+      value = value[field];
+    }
+
+    return value;
   }
 }
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -278,6 +278,53 @@ void main() {
     }));
   });
 
+  test('Where queries on nested objects should return the correct snapshots', () async {
+    final instance = MockFirestoreInstance();
+    final bookmarks = await instance
+        .collection('users')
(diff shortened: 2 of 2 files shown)


lib/src/mock_query.dart
```diff
@@ -90,7 +90,7 @@ class MockQuery extends Mock implements Query {
       List<dynamic> whereIn,
       bool isNull}) {
     _QueryOperation operation = (documents) => documents
-        .where((document) => _valueMatchesQuery(document[field],
+        .where((document) => _valueMatchesQuery(_getValueFromDocument(document, field),
```
atn832, line 93: Your `_getValueFromDocument` utility method is very similar to [util.dart's getSubPath](https://github.com/atn832/cloud_firestore_mocks/blob/45b6837fd614389adf30bf02e4c3b614a1ff1c49/lib/src/util.dart#L3), except that in your case, the separator is `'.'` instead of `'/'`, and yours will throw an exception if the path does not exist.

Instead of making your own method, could you tweak util.dart's getSubPath to accept an optional `separator` argument and reuse it? It'll remove the code duplication and handle missing paths for you.

test/mock_query_test.dart
```diff
@@ -278,6 +278,53 @@ void main() {
...
+        .listen(expectAsync1((QuerySnapshot snapshot) {
+      expect(snapshot.documents.length, equals(1));
+      expect(snapshot.documents.first.data['pager']['number'], equals(10));
+    }));
+
+    instance
+        .collection('users')
+        .document(uid)
+        .collection('bookmarks')
+        .where('pager.number', isGreaterThan: 9)
```
atn832, line 321: Awesome tests! Could you add a final test that checks that for nonexistent paths, it does not throw an exception, but simply returns an empty list? Some thing like:

```dart
 instance
        .collection('users')
        .document(uid)
        .collection('bookmarks')
        .where('pager.some.deep.path', isEqualTo: 10)
        .snapshots()
        .listen(expectAsync1((QuerySnapshot snapshot) {
      expect(snapshot.documents.length, equals(0));
    }));
```

lib/src/mock_query.dart
atn832: Also in order to use `getSubPath`, you'll probably need to pass `document.data` instead of document and cast the field to `String`: `getSubPath(document.data, field as String)`.

Btw is `Query.where` ever used with a non-string field? What's the expected behavior then?

atn832:
@C-Alexander do you need more guidance on how to polish your PR?

psygo:
This is great. Can we get it to pass? I kinda really need it.

atn832:
Thanks @C-Alexander for filing the bug and suggesting a fix. Eventually someone else submitted #158, so I am closing this PR.

Issue #83, opened 2020-04-28
MockDocumentReference.snapshots should return a deep copy


atn832:
At https://github.com/atn832/cloud_firestore_mocks/blob/38c30d477399648551a589ffe2fc6fd239b21a77/lib/src/mock_document_reference.dart#L137, we should return a deep copy like it's done in https://github.com/atn832/cloud_firestore_mocks/blob/38c30d477399648551a589ffe2fc6fd239b21a77/lib/src/mock_document_reference.dart#L120

PR #96, opened 2020-05-29
Always do a deep copy of the data when creating a MockDocumentSnapshot.

amatveyakin:
Fixes atn832/cloud_firestore_mocks#83

### DIFF
--- a/lib/src/mock_document_reference.dart
+++ b/lib/src/mock_document_reference.dart
@@ -117,7 +117,7 @@ class MockDocumentReference extends Mock implements DocumentReference {
   @override
   Future<DocumentSnapshot> get({Source source = Source.serverAndCache}) {
     return Future.value(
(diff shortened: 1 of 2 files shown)


amatveyakin:
I've chosen a path slightly different from what @atn832 suggested in https://github.com/atn832/cloud_firestore_mocks/issues/83#issue-607972679

I suggest to do `deepCopy` inside `MockDocumentSnapshot` constructor rather than when calling it. The later seems more error-prone to me. For example, the bug report doesn't mention mock_collection_reference.dart, which also needs to be updated. If we do `deepCopy` on the outside it would be easy to forget to do it again at some point in the future.

@atn832, what do you think?

atn832:
Yes, your approach is better. Thanks!

atn832:
It took a while, but I just published your fix to https://pub.dev/packages/cloud_firestore_mocks#045. Thank you again!

Issue #77, opened 2020-04-19
Real Firebase project prevents .where(arrayContains:) requests


atn832:
```
I/flutter (19814): 00:01 +0: Where: array-contains (Cloud Firestore) [E]
I/flutter (19814):   PlatformException(Error performing setData, PERMISSION_DENIED: Missing or insufficient permissions., null)
I/flutter (19814):   package:flutter/src/services/message_codecs.dart 569:7        StandardMethodCodec.decodeEnvelope
I/flutter (19814):   package:flutter/src/services/platform_channel.dart 321:33     MethodChannel.invokeMethod
I/flutter (19814):   ===== asynchronous gap ===========================
I/flutter (19814):   dart:async/zone.dart 1064:19                                  _CustomZone.registerBinaryCallback
I/flutter (19814):   dart:async-patch/async_patch.dart 82:23                       _asyncErrorWrapperHelper
I/flutter (19814):   package:test_api/src/backend/invoker.dart                     Invoker.waitForOutstandingCallbacks.<fn>
I/flutter (19814):   dart:async/zone.dart 1126:13                                  _rootRun
I/flutter (19814):   dart:async/zone.dart 1023:19                                  _CustomZone.run
I/flutter (19814):   dart:async/zone.dart 1518:10                                  _runZoned
I/flutter (19814):   dart:async/zone.dart 1465:12                                  runZoned
I/flutter (19814):   package:test_api/src/backend/invoker.dart 239:5               Invoker.waitForOutstandingCallbacks
I/flutter (19814):   package:test_api/src/backend/declarer.dart 169:33             Declarer.test.<fn>.<fn>
```

Perhaps we should create a Firebase project of our own and set up the right permissions.

PR #101, opened 2020-06-27
Support null values in orderBy

### LINKED ISSUE #66: orderBy throws when field is null
Not sure if this is expected behavior, but this test fails:
```
 test('my tests', () async {
     await instance.collection('usercourses').add({
       'started_at': Timestamp.fromDate(DateTime.now().subtract(Duration(days: 4))),
       'completed_at': Timestamp.fromDate(DateTime.now())
     });
     await instance.collection('usercourses').add({
       'started_at': Timestamp.fromDate(DateTime.now().subtract(Duration(days: 5))),
       'completed_at': null
     });

    var query = instance.collection('usercourses')
      .orderBy('completed_at');

    query.snapshots().listen(expectAsync1(
        (event) {
          expect(event.documents.length, greaterThan(0));
        },
      ));
  });
```
```
ERROR: NoSuchMethodError: The getter 'seconds' was called on null.
Receiver: null
Tried calling: seconds
dart:core                                                            Object.noSuchMethod
package:cloud_firestore_platform_interface/src/timestamp.dart 89:26  Timestamp.compareTo
package:cloud_firestore_mocks/src/mock_query.dart 51:32              MockQuery.orderBy.<fn>.<fn>
dart:collection                                                      ListMixin.sort
package:cloud_firestore_mocks/src/mock_query.dart 48:18              MockQuery.orderBy.<fn>
package:cloud_firestore_mocks/src/mock_query.dart 35:23              MockQuery.getDocuments
```
I would expect that there would still be results sorted from fields with values to fields that are `null`. I think it works like that with regular firebase queries as well.

atn832:
Fixes #66.

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -65,6 +65,16 @@ class MockQuery extends Mock implements Query {
       sortedList.sort((d1, d2) {
         final value1 = d1.data[field] as Comparable;
         final value2 = d2.data[field];
+        if (value1 == null && value2 == null) {
+          return 0;
+        }
+        // Return null values first.
+        if (value1 == null) {
+          return -1;
+        }
+        if (value2 == null) {
+          return 1;
+        }
         final compare = value1.compareTo(value2);
         return descending ? -compare : compare;
       });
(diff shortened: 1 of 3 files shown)


atn832:
Note that it fails for the real Firebase because of permission errors, like in #77.

Issue #19, opened 2020-02-25
Run tests automatically on new PRs


atn832:
and perhaps show that they pass on the master branch too.

suztomo:
https://github.com/atn832/cloud_firestore_mocks/pull/39#pullrequestreview-374891408

> I wonder if eventually, it would be possible to run the full battery of unit tests from cloud_firestore_mocks/test/* on the mock library and emulator/real firebase as part of #19.

Nice idea. The challenge is resetting the state of emulator and real firebase. For cloud_firestore_mocks, reseting its state is easy; just creating a MockFirestore instance. For emulator and real firebase, I don't know how to reset their state (collection and documents) via Dart.

atn832:
Reference docs:

* https://flutter.dev/docs/deployment/cd#other-services
* https://github.com/marketplace/actions/flutter-action

PR #104, opened 2020-06-27
add test workflow

sensuikan1973:
run unit test on PR/push.
this improve PR quality check, I think.

example: https://github.com/sensuikan1973/cloud_firestore_mocks/pull/1

regards,


.github/workflows/widget_test.yaml
```diff
@@ -0,0 +1,19 @@
...
+      - master
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
```
atn832, line 17: Looks like fetch-depth defaults to 1. Can you remove this line?

https://github.com/actions/checkout

.github/workflows/widget_test.yaml
```diff
@@ -0,0 +1,19 @@
+name: Unit_Test
```
atn832, line 1: Is it possible to use spaces in the name? Like "Unit tests". This way it'll look human readable when we add a status badge.

See https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

atn832: It looks like it's indeed possible: https://help.github.com/en/actions/configuring-and-managing-workflows/configuring-a-workflow#workflow-file-example

atn832:
Can you also rename the file to `unit-tests.yaml`? To use the same dash separated filenames as the GitHub documentation.

sensuikan1973:
fixed! 🚀

atn832:
Thank you @sensuikan1973, this is great!

Fixes #19.

PR #110, opened 2020-07-01
Mock Document Changes

feroult:
Expand `MockSnapshot` to support the `documentChanges` API.

This adds support for added documents only, as `fireSnapshotUpdate` won't fire for other events.

Regards


atn832:
Thank you for the fix @feroult !

atn832:
It took a while, but I finally published your fix at https://pub.dev/packages/cloud_firestore_mocks/changelog#0461. Thanks again!

Issue #111, opened 2020-07-05
snapshot Stream with where or orderBy doesn't work as Stream


gki:
I've found an issue the stream with `where` or `orderBy` will receive only one data **before starting to listen the stream**.

## Note
If I don't use any `where` or `orderBy`, the snapshot stream can receive data even if starting listen it before adding data.

## Test example
I checked with unit test code below for this issue.

```dart
  test('Continuous data receive via stream with where', () async {
    final instance = MockFirestoreInstance();

    instance
        .collection('messages')
        .where('archived', isEqualTo: false)
        .snapshots()
        .listen(expectAsync1((snapshot) {
          print(
              'In snapshot() before adding data. doc length ${snapshot.documents.length}');
          for (final d in snapshot.documents) {
            expect(d.data['content'],
                anyElement([equals('hello!'), equals('hola!')]));
          }
        }, count: 3)); // initial [], when add 'hello!' and when add 'hola!'.

    // this should be received.
    await instance.collection('messages').add({
      'content': 'hello!',
      'archived': false,
    });
    print('added hello!');

    // this should not be received because of archived == true.
    await instance.collection('messages').add({
      'content': 'bonjour!',
      'archived': true,
    });
    print('added bonjour!');

    // this should be received.
    await instance.collection('messages').add({
      'content': 'hola!',
      'archived': false,
    });
    print('added hola!');

    instance
        .collection('messages')
        .where('archived', isEqualTo: false)
        .snapshots()
        .listen((event) {
      event.documents.forEach((element) {
        print(
            'In snapshot() after added all data. ${element.documentID} ${element.data}');
      });
    });
  });
```

This test fails by timeout error and here is the output.
```
Continuous data receive via stream with where:

In snapshot() before adding data. doc length 0
added hello!
added bonjour!
added hola!
In snapshot() after added all data. lsEKCn5ufzUcK3fPNRIS {content: hello!, archived: false}
In snapshot() after added all data. R1aaWwDRPdplmE5PCRmb {content: hola!, archived: false}
ERROR: TimeoutException after 0:00:30.000000: Test timed out after 30 seconds. See https://pub.dev/packages/test#timeouts
dart:async  _startMicrotaskLoop
```

gki:
Hi @atn832,
In my situation, my PR #112 was not enough to notify date when updating/removing.
Please check #115 when you have time.

I've checked #115 update locally and it can support my situation as expected.
Thanks.

gki:
[...]

PR #115, opened 2020-07-19
Bugfix/notify when update doc

gki:
# Objective
For https://github.com/atn832/cloud_firestore_mocks/issues/111, notify updated/deleted data results to existing stream which is listening parent collection.

# Changes
- update unit test logic to check updating/deleting data.
- update `fireSnapshotUpdate()` method to notify date to "parent path" stream.


atn832:
Thank you for fixing this!

Issue #89, opened 2020-05-20
Missing support for FilePath.documentID in MockQuery.where and MockQuery.orderBy


nukotsuka:
```dart
final snapshot = await MockFirestoreInstance()
  .collection('collectionName')
  .where(FilePath.documentID, isEqualTo: 'documentID')
  .getDocuments();
```

```dart
final snapshot = await MockFirestoreInstance()
  .collection('collectionName')
  .orderBy(FilePath.documentID)
  .getDocuments();
```
Both of the above codes throw exception: `type 'FieldPath' is not a subtype of type 'String'`.
This is caused by `document[field]` with `dynamic field` which can be a [String] and a [FilePath].

PR #116, opened 2020-07-20
Add FieldPath.documentId support in MockQuery.where & MockQuery.orderBy

nukotsuka:
Fixes: #89

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -139,18 +139,24 @@ class MockQuery extends Mock implements Query {
       List<dynamic> arrayContainsAny,
       List<dynamic> whereIn,
       bool isNull}) {
-    _QueryOperation operation = (documents) => documents
-        .where((document) => _valueMatchesQuery(document[field],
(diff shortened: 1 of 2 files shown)


lib/src/mock_query.dart
```diff
@@ -93,8 +93,18 @@ class MockQuery extends Mock implements Query {
     return MockQuery(this, (documents) {
       final sortedList = List.of(documents);
       sortedList.sort((d1, d2) {
-        final value1 = d1.data[field] as Comparable;
-        final value2 = d2.data[field];
+        dynamic value1;
```
atn832, line 96: Is it possible to declare this as a Comparable instead of dynamic? If field is a String, we know it's a Comparable. If it's a document ID, it's a String, so it's also a Comparable.

atn832:
Anyway, I can reformat myself. Thanks for the work!

atn832:
Just published your fix to https://pub.dev/packages/cloud_firestore_mocks/changelog#0461. Thanks again!

PR #122, opened 2020-07-26
Dart format

sensuikan1973:
Because I don't want to consider format, I added dart format check workflow.

I you like, please consider merge.


atn832:
Very nice, thank you @sensuikan1973!

PR #124, opened 2020-08-07
Use different namespace for collections and documents' data

### LINKED ISSUE #90: setData without merge option overwrites sub-collections
I ran into this because I have a use case where I need to write documents to a sub-collection of a non-existent parent document before actually creating the parent document. What is happening is the sub-collection documents get overwritten when the parent document is finally set unless I specify `merge:true` on both the batch that commits the sub-collection documents and the `setData` that creates the parent document. That differs from the behavior of the SDK. In my case, its easy enough to just use `merge:true` option in the production code so this doesn't really block me but wanted to bring it to your attention. Will send a PR with the failing test.

dumbbell:
A document may contain a key and a sub-collection with the same name: they should not conflict.

By having separate namespaces for data and sub-collections, this also fixes #90: indeed, updating a document's data without setting the merge flag won't remove sub-collections.

One limitation is the `dump()` function of the `MockFirestoreInstance` class: because it rerturns a single tree, precedence is given to the sub-collection is there is a name conflict between a sub-collection and a key in the data.

Fixes #90.

### DIFF
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -14,6 +14,7 @@ import 'util.dart';
 
 class MockFirestoreInstance extends Mock implements Firestore {
   final _root = <String, dynamic>{};
+  final _docsData = <String, dynamic>{};
   final _snapshotStreamControllerRoot = <String, dynamic>{};
 
   /// Saved documents' full paths from root. For example:
@@ -29,7 +30,7 @@ class MockFirestoreInstance extends Mock implements Firestore {
     assert(segments.length % 2 == 1,
         'Invalid document reference. Collection references must have an odd number of segments');
     return MockCollectionReference(this, path, getSubpath(_root, path),
-        getSubpath(_snapshotStreamControllerRoot, path));
+        _docsData, getSubpath(_snapshotStreamControllerRoot, path));
   }
 
   @override
@@ -46,6 +47,7 @@ class MockFirestoreInstance extends Mock implements Firestore {
         path,
         documentId,
         getSubpath(_root, path),
+        _docsData,
         _root,
         getSubpath(_snapshotStreamControllerRoot, path));
   }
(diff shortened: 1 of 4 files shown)


lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -106,8 +108,18 @@ class MockFirestoreInstance extends Mock implements Firestore {
   }
 
   String dump() {
+    final copy = deepCopy(_root);
```
atn832, line 111: Can you add a comment to this loop? Something like:

```
// Place each document at the right place in the collection hierarchy.
```

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -106,8 +108,18 @@ class MockFirestoreInstance extends Mock implements Firestore {
...
+    for (var doc in _docsData.entries) {
+      var out = getSubpath(copy, doc.key);
```
atn832, line 113: Can you add intermediate variables to make it more readable? Like `final docPath = doc.key`, `final documentData = doc.value`, and rename `out` by something like `parentCollection`.

atn832: You can also make doc, entry and out `final`.

atn832:
Also can you write unit tests?

lib/src/cloud_firestore_mocks_base.dart
dumbbell: Done!

lib/src/cloud_firestore_mocks_base.dart
dumbbell: Done!

dumbbell:
Here is the new patch, with a testcase added.

atn832:
Thank you! I'll publish an update this weekend.

dumbbell:
Awesome, thank you!

atn832:
Your PR is now out at https://pub.dev/packages/cloud_firestore_mocks/changelog#047. Thanks again!

Issue #129, opened 2020-09-05
Transaction error


AndreaCostanzo1:
Good afternoon,
I'm facing an issue every time I try to run a transaction of any kind because I receive the following error (notice: using "normal" firebase, transactions work without problem):

```
type '_InternalLinkedHashMap<dynamic, dynamic>' is not a subtype of type 'Future<Null>?'

package:cloud_firestore_mocks/src/cloud_firestore_mocks_base.dart 88:5  MockFirestoreInstance.runTransaction
```

By diving a little bit deeper in the changes of the most recent version I noticed this change:

```dart
//old version
  Future<Map<String, dynamic>> runTransaction(
      TransactionHandler transactionHandler,
      {Duration timeout = const Duration(seconds: 5)}) async {

//new version
  Future<T> runTransaction<T>(TransactionHandler<T> transactionHandler,
      {Duration timeout = const Duration(seconds: 30)}) async {

```
When I saw this change I tried to specify the transaction type (see below) to see if the issue got solved. However, even if the error disappears, the transaction is not executed.

```dart
//specify dynamic
FirebaseFirestore.instance.runTransaction<dynamic>(/*code*/);
```
So I'm guessing if the issue is related to the latest update of the package

sensuikan1973:
The related diff of official plugin is https://github.com/FirebaseExtended/flutterfire/commit/27d574e620c3a95792682397b12e54e40abe572f#diff-4b03d320cd06661a52dd22772cf78abfL82-R181

Would you provide sample code ?
If you can, please provide test case like this **https://github.com/atn832/cloud_firestore_mocks/blob/66b043f566cd858db9670e8bc4e228fc2a167580/test/cloud_firestore_mocks_test.dart#L842**

AndreaCostanzo1:
I tried different cases and I found that the error happens only if there isn't a return statement or you return null, while, on the official firebase plugin, even if the return statement is suggested is not mandatory.

```dart
//Sample generating the issue
final FirebaseFirestore firestoreMock = MockFirestoreInstance();

  final String mockID= 'anduqw098eh12nesad2';
  final Map<String, dynamic> mockUser = {
    'id': mockID,
    'nickname': 'nick',
    'email': 'not@relevant.com',
  };

  test('transaction get and update', () async {
    //GIVEN: SETUP DB ADDING A MOCK USER
    DocumentReference mockUserReference=firestoreMock.collection('users').doc(mockID);
[...]

PR #127, opened 2020-08-30
[Migration] support cloud_firestore 0.14.x

sensuikan1973:
## Overview
support cloude_firestore 0.14 version.
From the version, we have to migrate. See: https://firebase.flutter.dev/docs/migration

This PR resolve https://github.com/atn832/cloud_firestore_mocks/issues/125.

Although **I made meaningful commit mesaages and separated commits based on migration step as far as possible**, diff is very big.

### important commits
1. https://github.com/sensuikan1973/cloud_firestore_mocks/pull/4/commits/bbfe14183328d3d08a21ab145981e81abbd8e2a4: update-firebase-plugins
2. https://github.com/sensuikan1973/cloud_firestore_mocks/pull/4/commits/510bf084c9efc6018ba66d38e26be522b8a3ae92: platform-setup
3. https://github.com/sensuikan1973/cloud_firestore_mocks/pull/4/commits/3957cdb658239447c807e79e776f2484324712d6: follow breaking/new changes
4. https://github.com/sensuikan1973/cloud_firestore_mocks/pull/4/commits/83fdb9c4d4f1aa788174be27bf2285661dc8f30e: fix ios gitignore
5. https://github.com/sensuikan1973/cloud_firestore_mocks/pull/4/commits/6c0d4978aea315a48a5f6ab6a7a462afc4a5537e: also example

### driver_test
I also ensured that driver_test and example runs.

When you run driver_test, some warning messages will appear. But, I don't follow that on this PR. Those warning have nothing to do with this migration task.
```sh
# driver_test. iOS log.
$ flutter driver -d XXX --driver=test_driver/cloud_firestore_test.dart
Your Flutter application is created using an older version of the Android embedding. It's being deprecated in favor of Android embedding v2. Follow the steps on https://flutter.dev/go/android-project-migration to migrate your project.
Starting application: lib/main.dart
Warning: Missing build name (CFBundleShortVersionString).
Warning: Missing build number (CFBundleVersion).
Action Required: You must set a build name and number in the pubspec.yaml file version field before submitting to the App Store.
Warning: Podfile is out of date
  This can cause issues if your application depends on plugins that do not support iOS.
  See https://flutter.dev/docs/development/packages-and-plugins/developing-packages#plugin-platforms for details.
  If you have local Podfile edits you would like to keep, see https://github.com/flutter/flutter/issues/45197 for instructions.
To regenerate the Podfile, run:
  rm ios/Podfile

Running Xcode build...                                                  
[...]

PR #130, opened 2020-09-05
fix runTransaction return type bug

sensuikan1973:
## Overview
resolve https://github.com/atn832/cloud_firestore_mocks/issues/129

### background
I changed Transaction in https://github.com/atn832/cloud_firestore_mocks/pull/127/files#diff-f88a1580752ecb4e5e77d48fd3ef7cfcL75.
The background is **https://github.com/FirebaseExtended/flutterfire/commit/27d574e620c3a95792682397b12e54e40abe572f#diff-4b03d320cd06661a52dd22772cf78abfL82-R181**

But, I also had to change the return value of runTransaction.

### DIFF
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -85,7 +85,7 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
       handlerResult.values.forEach(_validateTransactionReturnValue);
     }
 
-    return handlerResult ?? {};
+    return handlerResult;
   }
 
   /// Throws PlatformException when the value is not allowed as values of the
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -182,7 +182,7 @@ class _DummyTransaction implements Transaction {
   Transaction delete(DocumentReference documentReference) {
     _foundWrite = true;
     documentReference.delete();
-    return _DummyTransaction();
+    return this;
   }
 
   @override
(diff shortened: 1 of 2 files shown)


sensuikan1973:
@atn832 I would like you to review 🙏

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -85,7 +85,7 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
       handlerResult.values.forEach(_validateTransactionReturnValue);
     }
 
-    return handlerResult ?? {};
```
sensuikan1973, line 88: this is bug reason

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -85,7 +85,7 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
       handlerResult.values.forEach(_validateTransactionReturnValue);
```
atn832, line 85: Since https://github.com/FirebaseExtended/flutterfire/issues/1642 is now fixed. Is it safe to remove this loop entirely?

atn832: If so, then the function can `return transactionHandler(transaction);` directly.

atn832 (approved):
Thanks @sensuikan1973, I just have one question before merging.

lib/src/cloud_firestore_mocks_base.dart
sensuikan1973: oh... thanks 👍 
we can remove workaround code. So, I removed https://github.com/atn832/cloud_firestore_mocks/pull/130/commits/cbe0d656c4cb844da1a93391fb590cdba5cdc092

PR #133, opened 2020-10-24
fix dump with document reference

technixo:
In the previous version. If there contains any DocumentReference fields. call to `instance.dump()` will throw an error `Converting object to an encodable object failed: Instance of 'MockDocumentReference' `

This pull request is fixing it.

### DIFF
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -92,7 +92,11 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
         // property, the sub-category takes precedence, meaning the returned
         // json will not return that document property.
         if (!docCopy.containsKey(property.key)) {
-          docCopy[property.key] = property.value;
+          final propertyValue = property.value;
+          docCopy[property.key] = propertyValue is MockDocumentReference ? {
+            'type': 'DocumentReference',
+            'path': propertyValue.path,
+          } : propertyValue;
         }
       }
     }
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -20,6 +20,7 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
   /// Saved documents' full paths from root. For example:
   /// 'users/abc/friends/foo'
   final Set<String> _savedDocumentPaths = <String>{};
+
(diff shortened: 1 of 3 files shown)


atn832:
Thanks for contributing a PR. I'll review it tomorrow. In the meantime,  would you mind formatting the code?

https://flutter.dev/docs/development/tools/formatting

technixo:
We've reformatted the code

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -78,6 +79,24 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
     return await transactionHandler(transaction);
   }
 
+  dynamic deepFixDocumentReferenceProperty(dynamic propertyValue) {
+    if (propertyValue is MockDocumentReference) return propertyValue.toJson();
+
+    if (propertyValue is List) {
+      for (var i = 0; i < propertyValue.length; i++) {
+        propertyValue[i] = deepFixDocumentReferenceProperty(propertyValue[i]);
```
atn832, line 87: Looks like `propertyValue` comes from `_docsData`, so this will accidentally overwrite the actual document value.

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -78,6 +79,24 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
...
+      }
+    }
+    if (propertyValue is Map) {
+      for (var mapProperty in propertyValue.entries) {
+        propertyValue[mapProperty.key] =
```
atn832, line 92: Same here.

atn832 (changes requested):
Instead of recursively replacing doc.value in `dump`, have you tried modifying `myEncode` to support `MockDocumentReference`? Fixing `myEncode` seems like the proper way to implement this feature. Something like:

```dart
  } else if (item is MockDocumentReference) {
    return "DocumentReference(${item.path})";
  }
```

https://github.com/atn832/cloud_firestore_mocks/blob/e270d9cb0503997f023d3aa078afe599684dce39/lib/src/util.dart#L46

atn832:
Thanks for proposing the PR and writing exhaustive tests. Can you try modifying `myEncode` instead? The tests should still work fine after that.

atn832:
Feel free to reopen if you plan to fix it.

Issue #33, opened 2020-03-06
Stop extending Mock


atn832:
Extending `Mock` makes it possible to call Mockito methods like `when` even though that's not how the library should be used. See #26.

atn832:
Kind of blocked by https://github.com/dart-lang/sdk/issues/41147.

atn832:
Done at https://github.com/atn832/cloud_firestore_mocks/commit/5f1e4a7451eea75ee899ee20a399ed6d75759a67 and released at https://pub.dev/packages/cloud_firestore_mocks/changelog#083.

PR #141, opened 2021-03-04
changed dependencies to use firestore 1.0.0

Jouby:


Jouby:
I fixed all analyzer errors except this one : 
```  
 info • This class (or a class that this class inherits from) is marked as '@immutable', but one or more of its instance fields aren't
          final: Mock._givenName, Mock._givenHashCode, Mock._defaultResponse • lib/src/mock_collection_reference.dart:16:7 •
          must_be_immutable
```

The issue is about `MockCollectionReference` class which extends `MockQuery` from mockito package, with field not final.

And the `MockCollectionReference` class implements `CollectionReference` from cloud_firestore package  which is now `@immutable`.

atn832:
Thank you for updating the dependencies and cleaning up the warnings @Jouby!

We can't remove `implements CollectionReference` because the whole point of the class is to behave like a CollectionReference.

The warning is due to Mock's fields, which are not final. There are two ways to do it:

1. The quick and easy way, which also is the recommended way, is to add an annotation to ignore the error. See https://github.com/dart-lang/mockito/issues/200#issuecomment-511601012.
2. The more tedious way is to stop depending on Mock. Since we fake everything, we don't actually use Mock's mocking functionality anymore. But this is a larger change. It requires stubbing 6 methods on MockQuery. If we were to do it, it would be better to do it in another PR. See #33.

The 6 methods we would need to stub to do 2.:
image

So for this PR, would you mind adding the ignore annotation, as well as a comment to remind us to clean it up later, when Mockito fixes the issue? Something like this:

```dart
// Required until https://github.com/dart-lang/mockito/issues/200 is fixed.
// ignore: must_be_immutable
```

Jouby:
Thank for sharing the Mockito issue link. 

I have updated my PR by adding ignore annotation and a comment with mockito issue link. 👍

atn832:
I just published your update to https://pub.dev/packages/cloud_firestore_mocks/changelog#070. Thanks again!

Issue #142, opened 2021-03-05
Migrate to null safety


Jouby:
https://dart.dev/null-safety/migration-guide

```
Showing dependencies that are currently not opted in to null-safety.
[✗] indicates versions without null safety support.
[✓] indicates versions opting in to null safety.

Package Name  Current  Upgradable  Resolvable           Latest  

direct dependencies:
mockito       ✗4.1.4   ✗4.1.4      ✓5.0.0-nullsafety.7  ✓5.0.0  
quiver        ✗2.1.5   ✗2.1.5      ✓3.0.0               ✓3.0.0  

dev_dependencies: all support null safety.
```

atn832:
I am getting this error when upgrading Mockito and Quiver:

> Because every version of flutter_driver from sdk depends on crypto 2.1.5 and analyzer >=1.0.0 depends on crypto ^3.0.0, flutter_driver from sdk is incompatible with analyzer >=1.0.0.
> And because mockito 5.0.0 depends on analyzer ^1.0.0 and no versions of mockito match >5.0.0 <6.0.0, flutter_driver from sdk is incompatible with mockito ^5.0.0.

Looks like cloud_firestore is facing the same issue:

![image](https://user-images.githubusercontent.com/3399854/110227120-596ace00-7f38-11eb-9e73-aa19bea29f19.png)

So I could either

- temporarily remove the flutter_driver dependency so that we can migrate to null safety. Then I won't be able to run integration tests, and things are more likely to break.
- wait until flutter_driver fixes its dependencies, then I'll upgrade this library. null safety will wait a bit longer.

I'll take a few days to consider the options.

Btw it looks like flutter_driver already depends on analyzer 1.0.0 on master (https://github.com/flutter/flutter/blob/a0ba6464089f906fefbf610732b2c9a7b66a1260/packages/flutter_driver/pubspec.yaml#L29).  And there is some info about the tool update being released in version flutter 1.26 (https://github.com/flutter/flutter/pull/75370). I don't know what this means since Flutter 2.0.1 was published a few days ago, and flutter_driver doesn't seem to have been updated.

atn832:
Noted! I'll see what I can do this weekend.

torbenkeller:
@atn832 that would be really nice

KevinKozj:
Hey, any update on this one? We are pending on it before we can move forward on our development since our service depends on `mockito ^5.0.0` and we also need to use this lib for unit tests. Thanks

[...]

PR #143, opened 2021-03-08
update quiver to 3.0.0

AlexHartford:


atn832:
Thank you @AlexHartford. I'll release an update with your change this weekend, and possibly with #142.

PR #144, opened 2021-03-12
Added metadata data in documents mock

zariweyo:
In order to test code which use metadata, it's necessary to add metadata map to documents mockup

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -1,5 +1,6 @@
 import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:cloud_firestore_mocks/src/mock_document_reference.dart';
+import 'package:cloud_firestore_mocks/src/mock_snapshot_metadata.dart';
 import 'package:cloud_firestore_mocks/src/util.dart';
 import 'package:mockito/mockito.dart';
 
@@ -10,10 +11,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   final Map<String, dynamic> _document;
   final bool _exists;
   final MockDocumentReference _reference;
+  MockSnapshotMetadata _metadata;
 
   MockDocumentSnapshot(
-      this._reference, this._id, Map<String, dynamic> document, this._exists)
-      : _document = deepCopy(document);
+      this._reference, this._id, Map<String, dynamic> document, this._exists, {MockSnapshotMetadata metadata})
+      : _document = deepCopy(document){
+        if(metadata != null){
+          _metadata = metadata;
+        }else{
+          _metadata = MockSnapshotMetadata();
+        }
+      }
 
   @override
   String get id => _id;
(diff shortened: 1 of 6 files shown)


lib/src/mock_document_snapshot.dart
```diff
@@ -10,10 +11,18 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   final Map<String, dynamic> _document;
   final bool _exists;
   final MockDocumentReference _reference;
+  MockSnapshotMetadata _metadata;
 
   MockDocumentSnapshot(
-      this._reference, this._id, Map<String, dynamic> document, this._exists)
-      : _document = deepCopy(document);
+      this._reference, this._id, Map<String, dynamic> document, this._exists,
+      {MockSnapshotMetadata metadata})
```
atn832, line 18: Since none of the code currently passes metadata, would you mind simply initializing it like so: `_metadata = MockSnapshotMetadata()` at line 14?

I'd prefer to keep the code as simple as possible. If in the future, you plan to make a system to set up custom metadata, then we can make the constructor take the optional parameter in another PR.

https://github.com/atn832/cloud_firestore_mocks/search?q=MockDocumentSnapshot

lib/src/mock_snapshot_metadata.dart
```diff
@@ -0,0 +1,21 @@
...
+  MockSnapshotMetadata({
+    hasPendingWrites = false,
+    isFromCache = false,
+  }) {
+    _hasPendingWrites = hasPendingWrites;
+    _isFromCache = isFromCache;
+  }
+
+  @override
+  bool get hasPendingWrites => _hasPendingWrites;
```
atn832, line 17: Same, I'd prefer to simply write it as :

```dart
class MockSnapshotMetadata extends Mock implements SnapshotMetadata {
  @override
  bool get hasPendingWrites => false;

  @override
  bool get isFromCache => false;
}
```

Until someone implements custom metadata values.

atn832:
Thank you for the PR and for going as far as updating the changelog @zariweyo! Would you mind adding a unit test to test the code, and also demonstrate how one would use your feature?

lib/src/mock_document_snapshot.dart
zariweyo: Ok, I did this like you suggested. Thanks

lib/src/mock_snapshot_metadata.dart
zariweyo: I changed this to simply the method. In the future, maybe I'll send another PR to enable set metadata mocks. Thanks

zariweyo:
> Thank you for the PR and for going as far as updating the changelog @zariweyo! Would you mind adding a unit test to test the code, and also demonstrate how one would use your feature?

As you suggested, I add a test to check metadata.

atn832:
Thank you @zariweyo, I will release it at the same time as #145, some time this weekend.

PR #145, opened 2021-03-14
support null-safety

### LINKED ISSUE #142: Migrate to null safety

### LINKED ISSUE #146: Update dependency versions
To solve dependency conflicts, we should depend on the lastest version of [mockito](https://pub.dev/packages/mockito). Also, we should update the Cloud Firestore dependency.

atn832:
Support null-safety.

I had to temporarily disable integration tests by removing the dependency to flutter_driver for it to work. Otherwise I get dependency errors on `flutter pub get`:

```
[cloud_firestore_mocks] flutter pub get
Running "flutter pub get" in cloud_firestore_mocks...           
Because analyzer >=1.0.0 depends on args ^2.0.0 and every version of flutter_driver from sdk depends on args 1.6.0, analyzer >=1.0.0 is incompatible with flutter_driver from sdk.

And because mockito >=5.0.0 depends on analyzer ^1.0.0, flutter_driver from sdk is incompatible with mockito >=5.0.0.

So, because cloud_firestore_mocks depends on both mockito ^5.0.0 and flutter_driver any from sdk, version solving failed.
pub get failed (1; So, because cloud_firestore_mocks depends on both mockito ^5.0.0 and flutter_driver any from sdk, version solving failed.)
exit code 1
```

I will restore flutter_driver after flutter_driver has updated its dependencies.

Fixes #142. Fixes #146.

### DIFF
--- a/example/pubspec.yaml
+++ b/example/pubspec.yaml
@@ -2,7 +2,7 @@ name: firestore_example
 description: Demonstrates how to use the cloud_firestore_mocks plugin.
 
 environment:
-  sdk: ">=2.9.1 <3.0.0"
+  sdk: ">=2.12.0 <3.0.0"
 
 dependencies:
   flutter:
@@ -11,8 +11,6 @@ dependencies:
   firebase_core: ^1.0.0
 
 dev_dependencies:
-  flutter_driver:
-    sdk: flutter
   test: any
   cloud_firestore_mocks:
     path: ../
--- a/lib/src/mock_collection_reference.dart
+++ b/lib/src/mock_collection_reference.dart
@@ -5,6 +5,7 @@ import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:cloud_firestore_mocks/cloud_firestore_mocks.dart';
 import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_interface.dart';
 
+import 'mock_collection_reference_platform.dart';
 import 'mock_document_reference.dart';
 import 'mock_document_snapshot.dart';
 import 'mock_query.dart';
@@ -26,7 +27,8 @@ class MockCollectionReference extends MockQuery implements CollectionReference {
   final String _path;
 
   // ignore: unused_field
-  final CollectionReferencePlatform _delegate = null;
+  final CollectionReferencePlatform _delegate =
+      MockCollectionReferencePlatform();
 
   StreamController<QuerySnapshot> get snapshotStreamController {
     if (!snapshotStreamControllerRoot.containsKey(snapshotsStreamKey)) {
@@ -40,7 +42,7 @@ class MockCollectionReference extends MockQuery implements CollectionReference {
       this.snapshotStreamControllerRoot,
       {isCollectionGroup = false})
       : _isCollectionGroup = isCollectionGroup,
-        super();
+        super(null, null);
 
   @override
   FirebaseFirestore get firestore => _firestore;
@@ -49,7 +51,7 @@ class MockCollectionReference extends MockQuery implements CollectionReference {
   String get path => _path;
 
   @override
-  DocumentReference get parent {
+  DocumentReference? get parent {
     final segments = _path.split('/');
     final segmentLength = segments.length;
     if (segmentLength > 1) {
@@ -69,14 +71,13 @@ class MockCollectionReference extends MockQuery implements CollectionReference {
   }
 
   @override
-  Future<QuerySnapshot> get([GetOptions getOptions]) async {
+  Future<QuerySnapshot> get([GetOptions? getOptions]) async {
     var documents = <MockDocumentSnapshot>[];
     if (_isCollectionGroup) {
       documents = _buildDocumentsForCollectionGroup(root, []);
     } else {
       documents = root.entries.map((entry) {
-        MockDocumentReference documentReference =
-            _documentReference(_path, entry.key, root);
+        final documentReference = _documentReference(_path, entry.key, root);
         return MockDocumentSnapshot(
           documentReference,
           entry.key,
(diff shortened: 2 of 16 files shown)


example/pubspec.yaml
```diff
@@ -11,8 +11,6 @@ dependencies:
   firebase_core: ^1.0.0
 
 dev_dependencies:
-  flutter_driver:
```
atn832, line 14: Temporarily disable integration tests

lib/src/mock_collection_reference.dart
```diff
@@ -26,7 +27,8 @@ class MockCollectionReference extends MockQuery implements CollectionReference {
   final String _path;
 
   // ignore: unused_field
-  final CollectionReferencePlatform _delegate = null;
+  final CollectionReferencePlatform _delegate =
```
atn832, line 30: Cannot be null anymore since the field in Platform Interface is non-nullable, so I made a Mock.

lib/src/mock_document_reference.dart
```diff
@@ -36,7 +37,7 @@ class MockDocumentReference extends Mock implements DocumentReference {
       this.docsData, this.rootParent, this.snapshotStreamControllerRoot);
 
   // ignore: unused_field
-  final DocumentReferencePlatform _delegate = null;
+  final DocumentReferencePlatform _delegate = MockDocumentReferencePlatform();
```
atn832, line 40: Cannot be null anymore since the field in Platform Interface is non-nullable, so I made a Mock.

atn832:
Note: Analyzer and Unit tests fail only for the disabled integration tests and the example app tests, which also relies on flutter_driver. I will fix them in a future PR when I can safely restore flutter_driver.

suztomo:
Ack

lib/src/mock_query.dart
```diff
@@ -264,48 +265,55 @@ class QuerySnapshotStreamManager {
   }
 
   String _retrieveParentPath(MockQuery query) {
+    // In theory retrieveParentPath should stop at the collection reference.
+    // So _parentQuery can never be null.
+    if (query._parentQuery == null) {
+      throw 'Unexpected';
```
suztomo, line 271: Throwing a String is strange. Assert?

lib/src/mock_query.dart
```diff
@@ -264,48 +265,55 @@ class QuerySnapshotStreamManager {
...
 
-  StreamController<QuerySnapshot> getStreamController(Query query) {
+  StreamController<QuerySnapshot> getStreamController(MockQuery query) {
     final path = _retrieveParentPath(query);
     final pathCache = _streamCache[path];
+    // Before calling `getStreamController(query)`, one should have called
+    // `register(query)` beforehand, so pathCache should never be null.
     if (pathCache == null) {
-      return null;
+      throw 'Unexpected';
```
suztomo, line 306: Throwing a String is strange

lib/src/write_task.dart
```diff
@@ -9,8 +9,8 @@ enum WriteCommand {
 }
 
 class WriteTask {
-  WriteCommand command;
-  DocumentReference document;
-  Map<String, dynamic> data;
-  bool merge;
+  WriteCommand? command;
```
suztomo, line 12: Is it possible to have null here?
(I don't feel so)

lib/src/write_task.dart
```diff
@@ -9,8 +9,8 @@ enum WriteCommand {
...
+  DocumentReference? document;
```
suztomo, line 13: Is it possible to have null here?
(I don't feel so)

lib/src/write_task.dart
```diff
@@ -9,8 +9,8 @@ enum WriteCommand {
...
+  Map<String, dynamic>? data;
```
suztomo, line 14: I think this may be null (e.g., delete) Good.

suztomo (commented):
I haven't tried non-null-by-default in my app yet. Added comments on lines I felt strange.

lib/src/write_task.dart
atn832: You are right. I thought setting it as non-nullable would force me to define a constructor, but I can use the `late` keyword instead. It's much cleaner now, thank you! Making it non-nullable as it should be, allowed me to remove lots of unnecessary `!`:

image

lib/src/write_task.dart
atn832: Added a comment to explain when it can be null.

atn832:
Thanks for the review @suztomo, please take another look.

suztomo (approved):
Looks good to me, but the checks are failing.

atn832:
Thanks @suztomo, I'll restore the integration tests as soon as possible.

AlexHartford:
Do you plan to release a null-safe version soon? Seeing as this was merged, even without integration tests a dev release would suffice.

atn832:
I actually released this PR last week (https://github.com/atn832/cloud_firestore_mocks/issues/142#issuecomment-803519082), but the standard way of using Mockito isn't null-safe. For cloud_firestore_mocks to be soundly null-safe, we'd have to start using Mockito's alternate way which uses annotations to generate code. I've filed #149 to keep track of it.

PR #151, opened 2021-04-09
Add startAt/endAt cursors and limitToLast to MockQuery

maxluchterhand1:
Hey,

I needed the functionality of the Query.startAt/Query.endAt methods for some tests in my project.
I implemented them in a way which only simulates the cursor on a concrete match, that means

[0, 1, 2, 3, 4, 5, 7, 8, 9].startAt(5) = [5,7,8,9]

[0, 1, 2, 3, 4, 5, 7, 8, 9].startAt(6) = [0, 1, 2, 3, 4, 5, 7, 8, 9]

The behavior of endAt currently slightly differs from the actual Firestore behavior.
 
In this implementation, endAt makes the query end at the **first** value that matches the endAt value. In Firestore, it ends on the **last** value. This is gonna be addressed in #166.

Also, this implementation only supports exact matches. This is going to be addressed in #167.

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -165,6 +165,18 @@ class MockQuery extends Mock implements Query {
   }
 
   @override
+  Query limitToLast(int length) {
+    assert(
+      parameters['orderedBy'] is List && parameters['orderedBy'].isNotEmpty,
+      'You can only use limitToLast if at least one orderBy clause is specified.',
+    );
+    return MockQuery(
+      this,
+      (docs) => docs.sublist(max(0, docs.length - length), docs.length),
+    );
+  }
+
+  @override
   Query where(dynamic field,
       {dynamic isEqualTo,
       dynamic isNotEqualTo,
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -24,7 +24,11 @@ class MockQuery extends Mock implements Query {
   /// "orderBy". Null if this is a collection reference.
   final _QueryOperation? _operation;
 
-  MockQuery(this._parentQuery, this._operation);
+  MockQuery(this._parentQuery, this._operation)
+      : parameters = _parentQuery?.parameters ?? {};
+
+  @override
+  final Map<String, dynamic> parameters;
 
   // ignore: unused_field
   final QueryPlatform _delegate = MockQueryPlatform();
@@ -90,6 +94,8 @@ class MockQuery extends Mock implements Query {
 
   @override
   Query orderBy(dynamic field, {bool descending = false}) {
+    if (parameters['orderedBy'] == null) parameters['orderedBy'] = [];
+    parameters['orderedBy'].add(field);
     return MockQuery(this, (docs) {
       final sortedList = List.of(docs);
       sortedList.sort((d1, d2) {
@@ -119,6 +125,37 @@ class MockQuery extends Mock implements Query {
         return descending ? -compare : compare;
       });
       return sortedList;
+    });
+  }
+
+  @override
+  Query startAt(List<dynamic> values) =>
+      _cursorUtil(values, (docs, index) => docs.sublist(max(0, index)));
+
+  @override
+  Query endAt(List<dynamic> values) => _cursorUtil(values,
+      (docs, index) => docs.sublist(0, index == -1 ? docs.length : index));
+
+  Query _cursorUtil(
+    List<dynamic> values,
+    List<DocumentSnapshot> Function(List<DocumentSnapshot> docs, int index) f,
+  ) {
+    return MockQuery(this, (docs) {
+      assert(
+        parameters['orderedBy'] != null &&
+            parameters['orderedBy'].length >= values.length,
+        'You can only specify as many start values as there are orderBy filters.',
+      );
+      if (docs.isEmpty) return docs;
+
+      var res = List.of(docs);
+      for (var i = 0; i < values.length; i++) {
+        final index = docs.indexWhere(
+          (doc) => doc.data()?[parameters['orderedBy'][i]] == values[i],
+        );
+        res = f(docs, index);
+      }
+      return res;
     });
   }
 
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -142,7 +142,7 @@ class MockQuery extends Mock implements Query {
       values: values,
       f: (docs, index) => docs.sublist(
             0,
-            index == -1 ? docs.length : index,
+            index == -1 ? docs.length : index + 1,
           ));
 
   /// Utility function to avoid duplicate code for cursor query modifier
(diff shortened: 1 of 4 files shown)


atn832:
Thanks for your PR @maxluchterhand1. I'll take a look this weekend. In the meantime, would you mind writing unit tests for your new code?

lib/src/mock_query.dart
```diff
@@ -122,6 +128,37 @@ class MockQuery extends Mock implements Query {
...
 
+  @override
+  Query startAt(List<dynamic> values) =>
+      _cursorUtil(values, (docs, index) => docs.sublist(max(0, index)));
+
+  @override
+  Query endAt(List<dynamic> values) => _cursorUtil(values,
+      (docs, index) => docs.sublist(0, index == -1 ? docs.length : index));
+
+  Query _cursorUtil(
```
atn832, line 139: Would you mind documenting this function, in particular what the f argument is supposed to be?

lib/src/mock_query.dart
```diff
@@ -122,6 +128,37 @@ class MockQuery extends Mock implements Query {
...
+        'You can only specify as many start values as there are orderBy filters.',
+      );
+      if (docs.isEmpty) return docs;
+
+      var res = List.of(docs);
+      for (var i = 0; i < values.length; i++) {
+        final index = docs.indexWhere(
+          (doc) => doc.data()?[parameters['orderedBy'][i]] == values[i],
+        );
+        res = f(docs, index);
```
atn832, line 156: Should it be `f(docs,...` or `f(res,...`?

lib/src/mock_query.dart
atn832, line 139: It'd help make the code slightly more readable if you passed `List<String> orderByKeys` as a parameter to this function. Would you mind doing that, so that _cursorUtil doesn't access parameters directly?

lib/src/mock_query.dart
maxluchterhand1: Yea, I think it's supposed to be f(res,...) actually.

atn832:
I tried your code with the following tests:

```dart
  test('startAt, endAt', () async {
    await instance.collection('temperatures').add({'temperature': 3});
    await instance.collection('temperatures').add({'temperature': 5});
    await instance.collection('temperatures').add({'temperature': 7});
    await instance.collection('temperatures').add({'temperature': 2});

    // Prints out (2, 3, 5, 7).
    print(
        (await instance.collection('temperatures').orderBy('temperature').get())
            .docs
            .map((s) => s.get('temperature')));

    final aboveFive = await instance
        .collection('temperatures')
        .orderBy('temperature')
        .startAt([5]).get();
    expect(aboveFive.docs.map((s) => s.get('temperature')), equals([5, 7]));

    final aboveZero = await instance
        .collection('temperatures')
        .orderBy('temperature')
        .startAt([0]).get();
    expect(
        aboveZero.docs.map((s) => s.get('temperature')), equals([2, 3, 5, 7]));

    final underSix = await instance
        .collection('temperatures')
        .orderBy('temperature')
        .endAt([6]).get();
    // Also prints out (2, 3, 5, 7).
    print(underSix.docs.map((s) => s.get('temperature')));
    expect(underSix.docs.map((s) => s.get('temperature')), [2, 3, 5]);

    final underOne = await instance
        .collection('temperatures')
        .orderBy('temperature')
        .endAt([1]).get();
    // Also prints out (2, 3, 5, 7).
    print(underOne.docs.map((s) => s.get('temperature')));
    expect(underOne.docs.isEmpty, isTrue);
  });
```

startAt works but not endsAt. For example underSix throws:
> Expected: [2, 3, 5]
>  Actual: MappedListIterable<MockQueryDocumentSnapshot, dynamic>:[2, 3, 5, 7]
>   Which: at location [3] is MappedListIterable<MockQueryDocumentSnapshot, dynamic>:[2, 3, 5, 7] which longer than expected

 Would you mind adding these tests and a few more, perhaps with two properties, to make sure the code works fine?

test/mock_query_test.dart
```diff
@@ -825,4 +825,76 @@ void main() {
...
+
+    expect(snapshots.docs, hasLength(3));
+
+    snapshots = await baseQuery.startAt(['Springfield', 'Missouri']).get();
+
+    expect(snapshots.docs, hasLength(2));
+
+    snapshots = await baseQuery.endAt(['Springfield']).get();
+
+    expect(snapshots.docs, hasLength(2));
```
atn832, line 889: This doesn't match the actual Firestore. I got `{name: Los Angeles, state: California}, {name: Springfield, state: Massachusetts}, {name: Springfield, state: Missouri}, {name: Springfield, state: Wisconsin}` instead. It seems like they ended it at the last document that still matches the endAt value.

test/mock_query_test.dart
```diff
@@ -825,4 +825,76 @@ void main() {
...
+
+    // Since there is no Springfield, Florida in our docs, it should ignore the second orderBy value
+    snapshots = await baseQuery.endAt(['Springfield', 'Florida']).get();
+
+    expect(snapshots.docs, hasLength(2));
```
atn832, line 894: This doesn't match the actual Firestore. I got `{name: Los Angeles, state: California}`. It seems like it stopped before hitting {name: Springfield, state: Missouri}, because Florida < Missouri.

test/mock_query_test.dart
atn832: To prevent this PR from getting too complicated, I filed #166. Can you add a note here to explain it is different from Firestore, and that it is tracked at #166? We can tackle this edge case later.

test/mock_query_test.dart
atn832: As you reminded me last time, this is expected as per your implementation, so I filed #167. Can you leave a note here to mention it's different from Firestore, and that it will be fixed by #167? This way whoever resolves the ticket later will know this unit test should be fixed as well.

test/mock_query_test.dart
```diff
@@ -825,4 +825,76 @@ void main() {
...
+        instance.collection('cities').orderBy('name').orderBy('state');
+
+    var snapshots = await baseQuery.startAt(['Springfield']).get();
+
+    expect(snapshots.docs, hasLength(3));
+
+    // Since there is no Springfield, Florida in our docs, it should ignore the second orderBy value
+    snapshots = await baseQuery.startAt(['Springfield', 'Florida']).get();
+
+    expect(snapshots.docs, hasLength(3));
```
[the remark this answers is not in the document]
atn832: Although it's by chance since Floria happens to be alphabetically smaller than those three state names, so Firestore returns the same thing.

atn832 (approved):
Thanks for the detailed explanation and thorough unit tests. I only have minor comments about the tests. Would you mind resolving them? Then I'll merge your PR and publish an update.

maxluchterhand1:
Hey, thanks for your feedback. I just took care of everything you asked for.

test/mock_query_test.dart
```diff
@@ -8,6 +8,11 @@ import 'query_snapshot_matcher.dart';
 
 const uid = 'abc';
 
+extension ToData on List<QueryDocumentSnapshot> {
```
atn832, line 11: Beautiful

atn832:
Thanks again for the PR. I published it to https://pub.dev/packages/cloud_firestore_mocks/changelog#083.

Also I annotated some of your tests when they were different with Firestore for future reference at https://github.com/atn832/cloud_firestore_mocks/commit/39f0ae41ee7842bbf613d5f5eee1b826ad587f9d.

PR #148, opened 2021-03-21
fix compilation issues in integration tests

atn832:
By using dependency_overrides, I can avoid dependency conflicts from flutter_driver.

PR #153, opened 2021-04-13
Fix Query Snapshot To Return a Non Nullable Type

lyledean1:
Fix Query Snapshot that returns a non nullable type from this PR https://github.com/FirebaseExtended/flutterfire/pull/5476

I tested this by using in my pubspec.yaml file and works 
```
  cloud_firestore_mocks: 
    git:
      url: git@github.com:lyledean1/cloud_firestore_mocks.git
      ref: fix-mocks
```

### DIFF
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -181,7 +181,7 @@ void main() {
 
     final query = instance.collection('users').orderBy(FieldPath.documentId);
     query.snapshots().listen(expectAsync1(
-      (QuerySnapshot event) {
+      (event) {
         expect(event.docs.first.id, ('1'));
         expect(event.docs[1].id, ('2'));
         expect(event.docs[2].id, ('3'));
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -175,14 +175,13 @@ void main() {
 
   test('orderBy returns documents sorted by documentID', () async {
     final instance = MockFirestoreInstance();
-    await instance.collection('users').doc('3').set({});
-    await instance.collection('users').doc('2').set({});
-    await instance.collection('users').doc('1').set({});
+    await instance.collection('users').doc('3').set({'value': 3});
+    await instance.collection('users').doc('2').set({'value': 2});
+    await instance.collection('users').doc('1').set({'value': 1});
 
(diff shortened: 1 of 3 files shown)


AlexHartford:
This at least allows my tests to run, there are still a few more issues to tackle. Thanks for this.

lyledean1:
> This at least allows my tests to run, there are still a few more issues to tackle. Thanks for this.

No problem @AlexHartford , I've updated it so fmt + unit tests for package are passing. The example folder seems like it needs a bit of work

atn832:
> > This at least allows my tests to run, there are still a few more issues to tackle. Thanks for this.
> 
> No problem @AlexHartford , I've updated it so fmt + unit tests for package are passing. The example folder seems like it needs a bit of work

Don't worry about the example-test, I am working on them in #148. Also they are already broken in main, so this PR doesn't have to fix them.

lyledean1:
> > > This at least allows my tests to run, there are still a few more issues to tackle. Thanks for this.
> > 
> > 
> > No problem @AlexHartford , I've updated it so fmt + unit tests for package are passing. The example folder seems like it needs a bit of work
> 
> Don't worry about the example-test, I am working on them in #148. Also they are already broken in main, so this PR doesn't have to fix them.

Ok that makes sense @atn832 , the analyzer is not working either - but I see that's failing on master. This PR fixes the issues currently with running tests locally since FirebaseExtended/flutterfire#5476. Can this be merged?

atn832:
I need to catch up with the FlutterFire change. I'll review your code this weekend and get back to you then!

test/mock_query_test.dart
```diff
@@ -175,12 +175,11 @@ void main() {
 
   test('orderBy returns documents sorted by documentID', () async {
     final instance = MockFirestoreInstance();
-    await instance.collection('users').doc('3').set({});
-    await instance.collection('users').doc('2').set({});
-    await instance.collection('users').doc('1').set({});
+    await instance.collection('users').doc('3').set({'value': 3});
```
atn832, line 178: I believe this is due to a bug in MockDocumentReference. I've filed #156 to investigate and fix it later if necessary.

Issue #137, opened 2020-12-11
Where query doesn't work in nested objects


dfourcfive:
Hello guys first thank you for this package 
so basically i'm trying to get documents where a field in a field should be equal to something 
it doesn't throw any kind of error so instead it print nothing 
here's my demo code:

```
  final snapshot = await instance.collection('users');
await snapshot
      .where('address.country', isEqualTo: 'Canada')
      .get()
      .then((value) {
    if (value != null) {
           value.docs.forEach((element) {
        print(element.data());
      });
    } else {
      //this doesn't happen at all :)
      print('its null');
    }
  });
```

atn832:
This is now supported as of version 0.8.2. See https://pub.dev/packages/cloud_firestore_mocks/changelog#082.

PR #155, opened 2021-04-17
Fix MockDocumentSnapshot#data method signature and improve test usage

alex0711:
Hi @atn832,

thanks a lot for providing these mocks.
When i tried to use them in one of my projects i just stumbled upon a few issues:

* data() in DocumentSnapshot returns a non-null value - the mock has overridden the method with a nullable return type. Therefore tests couldn't be run at all.
* There's no reset() method to clear the state in a tests setUp method
* Queries where the where conditions key included a maps key were not working (see MockDocumentSnapshot#get(key))

### DIFF
--- a/lib/src/cloud_firestore_mocks_base.dart
+++ b/lib/src/cloud_firestore_mocks_base.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_interface.dart'
     as firestore_interface;
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:mockito/mockito.dart';
 
@@ -22,6 +23,14 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
   final Set<String> _savedDocumentPaths = <String>{};
   MockFirestoreInstance() {
     _setupFieldValueFactory();
+  }
+
+  /// Clears the state
+  @visibleForTesting
+  void reset() {
+    _root.clear();
+    _docsData.clear();
+    _snapshotStreamControllerRoot.clear();
   }
 
   @override
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -20,14 +20,19 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   String get id => _id;
 
   @override
-  dynamic get(dynamic key) => _document![key];
+  dynamic get(dynamic key) {
+    if (_isComposedKey(key)) {
+      return getComposedKeyValue(key);
+    }
+    return _document![key];
+  }
 
   @override
-  Map<String, dynamic>? data() {
+  Map<String, dynamic> data() {
     if (_exists) {
       return deepCopy(_document);
     } else {
-      return null;
+      return {};
     }
   }
 
@@ -39,4 +44,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
 
   @override
   SnapshotMetadata get metadata => _metadata;
+
+  bool _isComposedKey(String key) {
+    return key.contains('.');
+  }
+
+  dynamic getComposedKeyValue(String key) {
+    var groups = key.split('.');
+    dynamic value = _document!;
+    for (var group in groups) {
+      value = value[group];
+    }
+    return value;
+  }
 }
--- a/test/cloud_firestore_mocks_test.dart
+++ b/test/cloud_firestore_mocks_test.dart
@@ -310,7 +310,7 @@ void main() {
     expect(nested['k1'], 'v1');
   });
 
-  test('Nonexistent document should have null data', () async {
+  test('Nonexistent document should have empty data', () async {
     final nonExistentId = 'nonExistentId';
     final instance = MockFirestoreInstance();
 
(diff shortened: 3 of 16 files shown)


lib/src/mock_document_snapshot.dart
```diff
@@ -20,14 +20,19 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
...
+  dynamic get(dynamic key) {
+    if (_isComposedKey(key)) {
+      return getComposedKeyValue(key);
+    }
+    return _document![key];
+  }
 
   @override
-  Map<String, dynamic>? data() {
+  Map<String, dynamic> data() {
```
atn832, line 31: Actually, `DocumentSnapshot.data()` is still nullable, so we shouldn't change its signature here. See [DocumentSnapshot's documentation](https://pub.dev/documentation/cloud_firestore/latest/cloud_firestore/DocumentSnapshot-class.html#instance-methods).

It's `QueryDocumentSnapshot.data()` that was made to be non-nullable. For this issue, @lyledean1 kindly fixed the issue in #153.

test/cloud_firestore_mocks_test.dart
```diff
@@ -319,7 +319,7 @@ void main() {
     expect(snapshot1, isNotNull);
     expect(snapshot1.id, nonExistentId);
     // data field should be null before the document is saved
```
atn832, line 321: This unit test was introduced in #32 to match Firestore's behavior. It turns out that if you query a non-existent document, when you retrieve its `data()`, Firestore returns null. So we should continue to return null here.

lib/src/cloud_firestore_mocks_base.dart
```diff
@@ -24,6 +25,14 @@ class MockFirestoreInstance extends Mock implements FirebaseFirestore {
     _setupFieldValueFactory();
   }
 
+  /// Clears the state
+  @visibleForTesting
+  void reset() {
+    _root.clear();
```
atn832, line 31: In your tests, instead of clearing the state, would it be possible for you to re-instantiate a MockFirestoreInstance? It would be much safer than trying to recycle one instance. Recycling a MockFirestoreInstance is difficult. For example your implementation doesn't reset _savedDocumentPaths properly. Also the library does a bunch of things with listeners.

I'd prefer not to add a `reset()` method if possible.

lib/src/mock_document_snapshot.dart
```diff
@@ -39,4 +44,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
 
   @override
   SnapshotMetadata get metadata => _metadata;
+
+  bool _isComposedKey(String key) {
```
atn832, line 48: This function is nice! Later we could use it in other places like in https://github.com/atn832/cloud_firestore_mocks/blob/d8be0b9c9d522a0d78ebe2303bc5756c09e75951/lib/src/mock_document_reference.dart#L124 to make the code more readable.

Would you mind renaming this to `_isCompositeKey` for consistency with the rest of the codebase?

lib/src/mock_document_snapshot.dart
```diff
@@ -39,4 +44,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
...
+    return key.contains('.');
+  }
+
+  dynamic getComposedKeyValue(String key) {
```
atn832, line 52: I tried searching the Firestore doc and wasn't able to find the official term for it. So I don't know what the proper term is. But for better or for worse, we called those "composite keys" in https://github.com/atn832/cloud_firestore_mocks/blob/d8be0b9c9d522a0d78ebe2303bc5756c09e75951/lib/src/mock_document_reference.dart#L120 and in other places. Would you mind renaming this to `getCompositeKeyValue` for consistency?

lib/src/mock_document_snapshot.dart
```diff
@@ -39,4 +44,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
...
+    var groups = key.split('.');
```
atn832, line 53: This can be final. Would you mind rewriting it as:

```dart
final compositeKeyElements = key.split('.'); 
```

lib/src/mock_document_snapshot.dart
```diff
@@ -39,4 +44,17 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
...
+    dynamic value = _document!;
+    for (var group in groups) {
```
atn832, line 55: This can be final. Also, for consistency, would you mind rewriting it as:

```dart
for (final keyElement in compositeKeyElements) {
```

test/mock_query_test.dart
```diff
@@ -325,6 +325,31 @@ void main() {
         }));
   });
 
+  test('Where clause resolves composed keys', () async {
```
atn832, line 328: Very nice tests!

atn832:
@alex0711 thank you for contributing this pull request! As you wrote in the description, this PR is doing 3 things, so it'd have been better to create one PR for each issue. There are a few reasons for this. Some tasks are straightforward to review and merge, while some others might require longer discussions. It also allows you to name a PR to state exactly what it does.

Would you mind renaming this PR to "Support queries using composite keys" and reverting the changes related to reset and the `data` signature change? Either that or you can create a new PR if it makes it easier for you.

Very excited for this! This feature has long been requested  (#81, #137).

alex0711:
@atn832 Thanks a lot for the review - I'll create a new PR for the composed key feature and close this one.

Issue #140, opened 2021-02-24
Missing support for isNotEqualTo and isNull Query parameters


klisiewicz:
Having queries that utilize `isNotEqualTo` or `isNull`:
```dart
firestore.collection('document').where('status', isNotEqualTo: 'draft')
firestore.collection('document').where('publishedAt', isNull: true)
```

Results in a runtime error, since these operations are not supported:
```
package:cloud_firestore_mocks/src/mock_query.dart 243:5   MockQuery._valueMatchesQuery
package:cloud_firestore_mocks/src/mock_query.dart 149:18  MockQuery.where.<fn>.<fn>
dart:core                                                 Iterable.toList
package:cloud_firestore_mocks/src/mock_query.dart 159:12  MockQuery.where.<fn>
package:cloud_firestore_mocks/src/mock_query.dart 41:28   MockQuery.get

Unsupported
```

This makes some code impossible to test with `cloud_firestore_mocks` package.

atn832:
Released to https://pub.dev/packages/cloud_firestore_mocks/changelog#083.

PR #158, opened 2021-04-18
Support queries using composite keys

alex0711:
Hi @atn832,

this is the follow up of PR #155 which adds support for queries with composed keys.

When i tried writing some more tests, i noticed a few more minor issues:
* Queries with IsNull are not yet supported
* Queries with a composed key that does not exist (hence returns a null value) will fail as null is tried to cast to Comparable.
I have fixed them but i guess i should better put them in separate PRs as they are not only related to queries with composite keys?
Therefore some tests regarding the composite key feature are missing (query with a composite key that does not exist + query with isNull).

Regards,
Alex

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -20,7 +20,12 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
   String get id => _id;
 
   @override
-  dynamic get(dynamic key) => _document![key];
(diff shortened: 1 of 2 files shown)


atn832:
Thank you @alex0711! I just published your PR to https://pub.dev/packages/cloud_firestore_mocks/changelog#082.

That's right, these issues are better fixed in their own PRs. The missing isNull implementation was reported in #140. If you are willing to submit fixes for these, I'll happily review them!

PR #164, opened 2021-05-09
Fix null pointer exception

ming-chu:
There is a null pointer exception when query using `where()` cause with nested fields where the top-level name is missing.

e.g: `collectionReference.where('user.age', isEqualTo: 18);`
This query will failed if `user` is null.

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -54,6 +54,7 @@ class MockDocumentSnapshot extends Mock implements DocumentSnapshot {
     dynamic value = _document!;
     for (final keyElement in compositeKeyElements) {
       value = value[keyElement];
+      if (value == null) return null;
     }
     return value;
   }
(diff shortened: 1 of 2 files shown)


atn832:
Btw I just tried it on a test project, and it turns out Firestore throws an exception, instead of returning null:
image

So it'd be better if your PR simulated the same behavior.

My snippet:
```dart
FirebaseFirestore firestore = FirebaseFirestore.instance;
CollectionReference users = firestore.collection('users');
final docRef = await users.add({
  'full_name': "John Doe",
  'company': "Stokes and Sons",
  'age': 42,
  'preferences': {
    'hobbies': 3,
  },
});
final docSnapshot = await docRef.get();
print(docSnapshot.get('age'));
print(docSnapshot.get('preferences.hobbies'));
print(docSnapshot.get('preferences.favoritedrink'));
```

ming-chu:
@atn832 I added a test, please check. Hope this helps.

This is my test result before the fix.
image

atn832:
Oh that's what you were fixing. I get it now. I'll merge this. Thank you!

For the case I talked about, I filed #165.

Issue #175, opened 2021-05-20
Make `snapshots` support converters


atn832:
Right now it only works when calling `CollectionRef/DocumentRef/Query.get()`.

atn832:
Released to https://pub.dev/packages/fake_cloud_firestore/changelog#102.

PR #176, opened 2021-05-20
implemented convertedDoc.set and tested snapshots

atn832:


lib/src/mock_collection_reference.dart
```diff
@@ -81,16 +81,16 @@ class MockCollectionReference<T extends Object?> extends MockQuery<T>
 
   @override
   Future<QuerySnapshot<T>> get([GetOptions? options]) async {
-    List<Future<DocumentSnapshot<T>>> documents;
```
atn832, line 84: tiny cleanup

atn832:
Verifies some of #175.

It looks like right now, `snapshots` works only if one doesn't listen to both the unconverted and converted refs.

PR #187, opened 2021-09-08
fix comparisons with null

### LINKED ISSUE #186: missing key in map yields TypeError
If you try a where clause on a field of a map that does not exist, then it raises a `TypeError`:  "`type 'Null' is not a subtype of type 'Comparable<dynamic>'`".

If you try the same with the firebase emulator, it will not throw this error.

The Test:
```dart
final collectionID = "collection";

Future<bool> runTest(FirebaseFirestore firestore) async {

  await firestore.collection(collectionID).add({
    'mymap': {
      'someData': 2
    }
  });

  //Test for missing skill
  var query = firestore.collection(collectionID)
      .where('mymap.missingID', isGreaterThanOrEqualTo: 1);
  var qsnapshot = await query.get();
  return  (qsnapshot.size == 0); // should always be true
}
```

With Fake Firestore:
```dart
  bool queryEmpty = await runTest(FakeFirebaseFirestore()); // throws an TypeError
```

With the Emulator:
```dart

import 'package:cloud_firestore/cloud_firestore.dart';
import 'package:firebase_core/firebase_core.dart';
import 'package:flutter/material.dart';

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  await Firebase.initializeApp();
  // using emulator (set up in index.html)

  bool queryEmpty = await runTest(FirebaseFirestore.instance);

  runApp(MaterialApp(home: Text("With emulator: $queryEmpty",))); // this works for me
}
```

I believe the error is because `_valueMatchesQuery()` from `MockQuery` does not check if `value` contains `null`. (Except if the query is a "isNull"-query). If the method only needs to return `false` if it isnt a comparison with null, then I would try to change this myself and contribute it.

fkoch-tgm:
I believe this should resolve #186

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -242,6 +242,8 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
     } else if (isNull != null) {
       final valueIsNull = value == null;
       return isNull ? valueIsNull : !valueIsNull;
+    } else if (value == null) {
+      return false;
     } else if (isGreaterThan != null) {
       Comparable fieldValue = value;
       if (isGreaterThan is DateTime) {
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -306,6 +306,22 @@ void main() {
         .listen(expectAsync1((QuerySnapshot snapshot) {
       expect(snapshot.docs.length, equals(0));
     }));
+
+    instance
+        .collection('contestants')
+        .where('skills.swimming', isEqualTo: '1')
+        .snapshots()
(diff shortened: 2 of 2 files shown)


atn832:
Thank you for the fix! I'll publish a new version with your fix by tomorrow.

Btw, the bug happened for non nested fields too, so I added a simpler test. I also modified your condition to more explicitly handle comparisons at [941feb1245b1e26617dd056509952f9ed5dcaca9](https://github.com/atn832/fake_cloud_firestore/commit/941feb1245b1e26617dd056509952f9ed5dcaca9).

atn832:
Your fix is out. https://pub.dev/packages/fake_cloud_firestore/changelog#120

Issue #191, opened 2021-11-05
Removing a document invokes `fromFirestore` from the `withConverter`


wesleyfuchter:
Hey there!

Just want to say this package is helping a lot during my learning on how to test a Firestore service.

I'm using `fake_cloud_firestore` to unit test a service class that removes a document. Here is the removing code:

```
 @override
  Future<void> remove(String id) async {
      await _tasksCollection.doc(id).delete();
  }
```

I'm using a converter to convert from JSON to my entity `Task` as:

```
class TaskFirestoreAdapter implements FirestoreAdapter<Task> {
  final _logger = Logger();

  @override
  Task fromFirestore(DocumentSnapshot<Map<String, dynamic>> document, _) {
    _logger.i('Parsing firestore document ${document.id} to task');    
    return Task.fromJson(document.id, document.data()!);
  }

  @override
  Map<String, Object?> toFirestore(Task model, _) {
    _logger.i('Parsing task ${model.id} to firestore document');
    return model.toJson();
  }
}
```

And here is my test:

```
test('Should remove a given task with its id', () async {
  final taskService = getIt<TaskService>();
  final query = await instance.collection('tasks').where('title', isEqualTo: 'Task 1').get();
  expect(query.docs.isEmpty, isFalse);
  await taskService.remove(query.docs.first.id);
  final queryAfterRemoving = await instance.collection('tasks').where('title', isEqualTo: 'Task 1').get();
  expect(queryAfterRemoving.docs.isEmpty, isTrue);
});
```

When I run this test I get the following error:

```
flutter test test/task_service_test.dart --name "Should remove a given task with its id"
00:02 +0: TaskService Should remove a given task with its id
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
00:02 +0 -1: TaskService Should remove a given task with its id [E]
  Null check operator used on a null value
  package:flutter_testing_experiment/src/core/services/task_firestore_adapter.dart 13:54  TaskFirestoreAdapter.fromFirestore
  package:fake_cloud_firestore/src/mock_document_reference.dart 204:52                    MockDocumentReference.get
  package:fake_cloud_firestore/src/mock_document_reference.dart 233:40                    MockDocumentReference.fireSnapshotUpdate
  package:fake_cloud_firestore/src/mock_document_reference.dart 219:5                     MockDocumentReference.delete
  package:flutter_testing_experiment/src/core/services/task_service.dart 49:36            TaskService.remove
[...]

PR #192, opened 2021-11-05
On removing document do not call the converter if document doesn't exists.

wesleyfuchter:
When deleting a document the method `fireSnapshotUpdate` is called triggering a call to the converter (after some levels). The problem on calling the converter with the document as `_exists()` being false is that will trigger a null problem inside the converter (that requires non null values per Firebase lib). 

After looking the code I didn't understand the reasons on calling the `get()` method and subsequently the converter for a remove operation, so I added this verification before. All the tests are passing, so I hope this solves the issue #191.

### DIFF
--- a/lib/src/mock_document_reference.dart
+++ b/lib/src/mock_document_reference.dart
@@ -217,7 +217,9 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
     rootParent.remove(id);
     _firestore.removeSavedDocument(path);
     QuerySnapshotStreamManager().fireSnapshotUpdate(firestore, path);
-    fireSnapshotUpdate();
+    if (_exists()) {      
+      fireSnapshotUpdate();
+    }    
     return Future.value();
   }
 


lib/src/mock_document_reference.dart
```diff
@@ -217,7 +217,9 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
     rootParent.remove(id);
     _firestore.removeSavedDocument(path);
     QuerySnapshotStreamManager().fireSnapshotUpdate(firestore, path);
-    fireSnapshotUpdate();
+    if (_exists()) {
```
atn832, line 220: On line 218, since we remove the saved document, I believe it won't exist. So perhaps we can just remove the fireSnapshotUpdate altogether?

atn832 (changes requested):
Thanks for catching this. It must have been hard to debug! Would you mind removing fireSnapshotUpdate altogether?

atn832:
Anyway, I'll just modify it myself since it's a small change. Thanks!

atn832:
I remove the call here: https://github.com/atn832/fake_cloud_firestore/commit/d1b42de97099362c44173004cfa34602777e894f. I will release an update by tomorrow.

atn832:
Your fix is out. https://pub.dev/packages/fake_cloud_firestore/changelog#120

lib/src/mock_document_reference.dart
wesleyfuchter: Sorry, I missed this. Thanks for pointing that out and doing the change. 💯

PR #197, opened 2022-01-10
Implement missing query modifiers and fixes an issue with document snapshots

### LINKED ISSUE #211: Support for endAtDocument, endBefore, endBeforeDocument and startAtDocument
I found this library very useful but it was just missing some query modifiers: endAtDocument, endBefore, endBeforeDocument and startAtDocument
And I needed them in my tests so I implemented them in this pr #197

Andrew-Bekhiet:
Implements endAtDocument, endBefore, endBeforeDocument and startAtDocument
Fixes document.snapshots() not emitting first event when listened to
Closes #211


lib/src/mock_query.dart
```diff
@@ -323,21 +323,40 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
...
-  Query<T> endAtDocument(DocumentSnapshot documentSnapshot) {
-    // TODO: implement endAtDocument
-    throw UnimplementedError();
+  Query<T> endAtDocument(DocumentSnapshot snapshot) {
+    return MockQuery(this, (docs) {
+      final index = docs.indexWhere((doc) {
+        return doc.id == snapshot.id;
+      });
+
+      return docs.sublist(0, index == -1 ? docs.length : index + 1);
```
atn832, line 332: Would you mind adding a comment that says something like:

> Return all documents up to snapshot. If not found, return everything.

lib/src/mock_query.dart
```diff
@@ -323,21 +323,40 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
...
-  Query<T> endBefore(List values) {
-    // TODO: implement endBefore
-    throw UnimplementedError();
-  }
+  Query<T> endBefore(List values) => _cursorUtil(
+      orderByKeys: parameters['orderedBy'] ?? [],
+      values: values,
+      f: (docs, index, exactMatch) => docs.sublist(
+            0,
+            index == -1 ? docs.length - 1 : index,
```
atn832, line 342: Would you mind describing what this does as well, something like:

> Return documents up to values. If not found, return everything instead.

lib/src/mock_query.dart
```diff
@@ -323,21 +323,40 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
...
-    // TODO: implement endBeforeDocument
-    throw UnimplementedError();
+  Query<T> endBeforeDocument(DocumentSnapshot snapshot) {
+    return MockQuery(this, (docs) {
+      final index = docs.indexWhere((doc) {
+        return doc.id == snapshot.id;
+      });
+
+      if (index == -1) {
+        throw PlatformException(
```
atn832, line 353: How come in the other queries, when it's not found, it returns everything. But for this one query, it throws an exception? Is that the behavior in Firestore?

test/fake_cloud_firestore_test.dart
```diff
@@ -135,6 +139,88 @@ void main() {
...
+                {},
+              ),
+            ),
+            predicate<DocumentSnapshot>((s) => !s.exists),
+          ],
+        ),
+      );
+
+      //Await the first snapshot to start the stream
+      await allSnapshots.first;
```
atn832, line 173: I wonder if this workaround is still needed, since you introduced `_getSync` and `stream.startWith`. But you don't have to check. I will check and clean up later if needed.

atn832:
Your implementations look good. I left a few questions. Would you mind adding unit tests for each case that you implemented? This it to make sure everything works as intended, and that no one accidentally breaks your code in the future.

atn832 (changes requested):

lib/src/mock_query.dart
Andrew-Bekhiet: I copied the code from ```startAfterDocument``` and tweaked it so that they have the same logic, but I honestly don't know how firestore behaves in this case

Andrew-Bekhiet:
Added comments and test cases

test/mock_query_test.dart
```diff
@@ -576,30 +576,129 @@ void main() {
...
+
+      final documentSnapshot =
+          await instance.collection('messages').doc(uid).get();
+
+      final snapshots = await instance
+          .collection('messages')
+          .startAtDocument(documentSnapshot)
+          .get();
+
+      expect(snapshots.docs, hasLength(3));
```
atn832, line 624: I would have mapped each doc to their Username and expected the actual username values. That would have made the tests slightly more readable. But it's ok, your tests do the job too.

atn832 (approved):

lib/src/mock_query.dart
atn832: Alright. It's ok, we can check the actual behavior and refine later.

atn832:
Last remark. Since the fix with document.snapshots() is unrelated to queries, it would have been better to make it a separate PR. It simplifies code review and helps us make each PR cleaner. For example, had `document.snapshots()` been its own PR, we could more easily have refined the affected unit tests I mentioned at https://github.com/atn832/fake_cloud_firestore/pull/197#discussion_r829615362 without overcomplicating this PR.

atn832:
I just published your changes to https://pub.dev/packages/fake_cloud_firestore/changelog#122. Thanks again for your help!

PR #202, opened 2022-02-08
check if document exists before converting

cedvdb:


atn832:
Thanks for proposing the fix. I literally just saw this after working on #191 and submitting #209, so I won't merge your PR. I'll still leave credits to you in the changelogs!

PR #203, opened 2022-02-08
fix invalid range

cedvdb:


cedvdb:
this pull request is not correct but the original code does not work and its intent is unclear...

PR #205, opened 2022-02-17
add FieldPath support to doc.get()

jpangburn:
The Firestore documentation at [https://googleapis.dev/nodejs/firestore/latest/DocumentSnapshot.html#get](https://googleapis.dev/nodejs/firestore/latest/DocumentSnapshot.html#get) shows you should be able to do gets like this `documentSnapshot.get('a.b')` to access child fields directly.  Here's a full example from that page:

> documentRef.set({ a: { b: 'c' }}).then(() => {
>   return documentRef.get();
> }).then(documentSnapshot => {
>   let field = documentSnapshot.get('a.b');
>   console.log(`Retrieved field value: ${field}`);
> });

This currently works in fake_cloud_firestore.  What's missing is that the spec says you should be able to use a FieldPath object instead of a String as the key.  The reason for this is that a field name can have dots in it.  So for example you may need to use `FieldPath(['a','I.have.dots'])` in order to access such a field because obviously `'a.I.have.dots'` would be trying to lookup a different field than the one you intended.

This PR provides unit tests to test both of these scenarios (although the first scenario was working already, I didn't find a test for it).  The second test may illustrate what this is trying to solve better than my explanation above.

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -54,12 +54,21 @@ class MockDocumentSnapshot<T extends Object?> implements DocumentSnapshot<T> {
   @override
   SnapshotMetadata get metadata => _metadata;
 
-  bool _isCompositeKey(String key) {
-    return key.contains('.');
+  bool _isCompositeKey(dynamic key) {
+    if (key is String) {
+      return key.contains('.');
+    } else if (key is FieldPath) {
+      return true;
+    } else {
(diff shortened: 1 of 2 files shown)


atn832:
Very nice. Also, thanks for also implementing unit tests for the pre-existing support of "a.b" key strings!

jpangburn:
Thank you for maintaining this very helpful project!  Glad I could help a little :-)

PR #206, opened 2022-02-17
Fix notequalto query

jpangburn:
The existing notEqualTo query implementation ignores the specification at [https://firebase.google.com/docs/firestore/query-data/queries#not_equal_](https://firebase.google.com/docs/firestore/query-data/queries#not_equal_) part that says 

> Use the not equal (!=) operator to return documents where the given field exists and does not match the comparison value

The part that's not implemented is `the given field exists`.  This PR resolves that issue.  It also allows the use of FieldPath values to specify the field instead of just String values (which the real Firestore supports).

This PR also provides test to verify fixed functionality, including the use of FieldPath.  It requires the previous PR I created #205 in order to have FieldPath support.

### DIFF
--- a/lib/src/mock_document_snapshot.dart
+++ b/lib/src/mock_document_snapshot.dart
@@ -54,12 +54,21 @@ class MockDocumentSnapshot<T extends Object?> implements DocumentSnapshot<T> {
   @override
   SnapshotMetadata get metadata => _metadata;
 
-  bool _isCompositeKey(String key) {
-    return key.contains('.');
+  bool _isCompositeKey(dynamic key) {
+    if (key is String) {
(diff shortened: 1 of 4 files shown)


atn832:
Thank you for making `notEqualTo` stick to the specs. I'll merge your PR, then wait a bit for https://github.com/atn832/fake_cloud_firestore/pull/197 so I can release both changes at the same time.

atn832 (approved):

jpangburn:
You're welcome, thanks again for maintaining this project!

atn832:
My pleasure. Just published your changes to version 1.2.2. https://pub.dev/packages/fake_cloud_firestore/changelog#122

PR #215, opened 2022-03-26
Make some tests that should fail, actually fail.

cedvdb:
Make some tests that should fail, actually fail. The fact that the tests were engineered to succeed is concerning.

From those failing tests we can see that `orderBy` is failing, `endAt` and `startAt`

Side effect of closing https://github.com/atn832/fake_cloud_firestore/issues/204

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -115,7 +115,7 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
       values: values,
       f: (docs, index, exactMatch) => docs.sublist(
             0,
-            index == -1 ? docs.length : index + 1,
+            index,
           ));
 
   /// Utility function to avoid duplicate code for cursor query modifier
@@ -150,7 +150,7 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
       if (docs.isEmpty) return docs;
 
       var res;
-      var found = false;
+      var exactMatch = false;
       for (var i = 0; i < values.length; i++) {
         var sublist = docs.sublist(res ?? 0);
         final keyName = orderByKeys[i];
@@ -163,11 +163,12 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
               final docValue = doc.get(keyName);
               return docValue.compareTo(searchedValue) == -1;
             });
-        found = sublist[index].get(keyName) == searchedValue;
+        exactMatch = index < sublist.length &&
+            sublist[index].get(keyName) == searchedValue;
         res = res == null ? index : res + index;
       }
 
-      return f(docs, res ?? -1, found);
+      return f(docs, res ?? -1, exactMatch);
     });
   }
 
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -1127,7 +1127,7 @@ void main() {
     ]);
   });
 
-  test('startAt/endAt', () async {
+  test('orderBy', () async {
     final instance = FakeFirebaseFirestore();
 
     await instance.collection('cities').doc().set({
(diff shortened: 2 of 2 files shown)


test/mock_query_test.dart
```diff
@@ -1232,10 +1359,13 @@ void main() {
...
       },
+      {
+        'name': 'Springfield',
+        'state': 'Missouri',
+      },
     ]);
 
-    snapshots = await baseQuery.endAt(['Springfield', 'Missouri']).get();
-
+    snapshots = await baseQuery.endAt(['Wellington']).get();
```
atn832, line 1368: Should it return everything because Wellington is alphabetically grater than Washington, or because "Wellington" as a value does not exist, and so it returns everything?

test/mock_query_test.dart
```diff
@@ -1186,6 +1256,10 @@ void main() {
         'name': 'Springfield',
         'state': 'Wisconsin',
       },
+      {
+        'name': 'Washington',
```
atn832, line 1260: Nice catch, I wonder why we missed this.

test/mock_query_test.dart
```diff
@@ -1150,12 +1150,79 @@ void main() {
...
+    await instance.collection('cities').doc().set({
+      'name': 'Washington',
+      'state': 'Washington',
+    });
+
     final baseQuery =
         instance.collection('cities').orderBy('name').orderBy('state');
 
-    var snapshots = await baseQuery.startAt(['Springfield']).get();
+    var snapshots = await baseQuery.startAt(['Los Angeles']).get();
```
atn832, line 1219: What made you want to change the value here? Now we won't know if the test passes because startAt returned everything, or because it really matched with the value.

test/mock_query_test.dart
atn832: If it's the latter, it'd be nice to add a comment to explain it, and perhaps change Wellington to some dummy value so that it is obvious why the query returns those results.

lib/src/mock_query.dart
```diff
@@ -115,7 +115,7 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
       values: values,
       f: (docs, index, exactMatch) => docs.sublist(
             0,
-            index == -1 ? docs.length : index + 1,
```
atn832, line 118: I don't know what the real Firestore behavior is, but in our previous implementation, if the endAt value wasn't found, we would return everything. With your change, now we will return empty. Is that what you are trying to achieve?

test/mock_query_test.dart
cedvdb: I think I did it to be sure it returned everything when the index was 0. I don't know exactly. I will revert this.

lib/src/mock_query.dart
cedvdb: I will revert this, although the next PR removes this

test/mock_query_test.dart
cedvdb: Wellington is alphabetically greater so it returns everything that is under it. Do you want me to rename wellington to something clearer ? Do you have a variable name in mind ?

atn832: No, Wellington is fine. I just wasn't sure what the expected behavior was. Since edge cases with queries can get complicated, perhaps you can add a comment to explain that if the value is greater than all others, we expect everything to be returned. Thank you!

lib/src/mock_query.dart
cedvdb: Is it okay if I don't revert this as the next pull request essentially remove this ?  I'm afraid to have to rebase everything after this..

cedvdb:
I've added comments but there is still one of your comment I did not address. Please let me know if my response sounds okay so we can move forward with the next pull requests.

lib/src/mock_query.dart
atn832: Sure, let's keep it simple for you.

Issue #217, opened 2022-03-26
Nested updates are merged


cedvdb:
updating a nested maps merges the map instead of replacing it like firestore does

PR #216, opened 2022-03-26
Fix ordered by

cedvdb:
Make tests introduced in https://github.com/atn832/fake_cloud_firestore/pull/215 pass.

PR #218, opened 2022-03-26
Update fix

cedvdb:


lib/src/mock_query.dart
```diff
@@ -71,52 +71,100 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
...
-          return -1;
-        }
-        if (value2 == null) {
-          return 1;
-        }
-        final compare = value1.compareTo(value2);
-        return descending ? -compare : compare;
-      });
+      final fields = (parameters['orderedBy'] ?? []);
+      for (var index = 0; index < fields.length; index++) {
```
atn832, line 75: Is this necessary to fix 217? Can you explain what needed rework?

test/mock_query_test.dart
```diff
@@ -1249,24 +1379,32 @@ void main() {
...
-        .withConverter(fromFirestore: from, toFirestore: to);
-    await moviesCollection.add(Movie()..title = 'A long time ago');
-    await moviesCollection.add(Movie()..title = 'Robot from the future');
-    final searchResults =
-        await moviesCollection.orderBy('title').startAt(['R']).get();
-    expect(searchResults.size, equals(1));
-    final movieFound = searchResults.docs.first.data();
-    expect(movieFound.title, equals('Robot from the future'));
-  });
+  // test('startAt with converters', () async {
```
atn832, line 1393: Would you mind restoring the tests?

atn832:
It is not a good idea to chain PRs for unrelated issues. Since this PR includes code from the previous PRs, one can't easily distinguish which code is relevant to the issue. This makes reviewing the code much harder.

I've also made cleanups in #225 after merging #216, so this PR might get merge conflicts.

It might be easier for you to close this PR and make a brand new one to fix #217.

atn832 (changes requested):

lib/src/mock_query.dart
cedvdb: This PR is from 3 months ago and the code is not exactly straight forward at every turn. This might be from one of the PR this one was chained onto..

cedvdb:
> It is not a good idea to chain PRs for unrelated issues.

Agreed. I was trying to resolve a specific problems that affected me personally and applied every fix in succession to fix my problem. This is not exactly a good excuse but here we are.

lib/src/mock_query.dart
atn832: Thank you! And sorry for taking so long to review. It makes us forget the context. That's why I merged #216 as is, since it fixed the issue, then cleaned up the code myself.

cedvdb: For info I'm really busy right now. This is still on my mind, but I have priorities, I don't think I will be able to come back to this in the immediate future.

PR #226, opened 2022-06-24
Fix bug with replacing non-map values with map

Sapphirewhale:
Hi all, this is my first time contributing to a public codebase, so please let me know if there's anything out of place here.

# Problem: 
Previously the code worked under the assumption that if the new value for a given key was a map, and the key existed in the document, then the existing value was also a map. This is not always the case, as a value could be updated from an int to a map e.g. Admin.permissions = null => Admin.permissions = {deleteUser: False, createUser:True...}

In this case an unhandled TypeError is thrown and interrupts the update call.

# Solution:
This fix updates the code which identifies whether a given key already exists in the document for a new value of type map, and has it also check whether the existing value is of type Map. If either of these conditions is false, the key is assigned a blank Map as a value, before applying the new value.

Unit tests all passed, and this fixed the issue in the project where I identified it, so hopefully everything looks good here. Please let me know if there's anything in the code that should be changed, I'm eager for feedback, and thank you so much for sharing this plugin! You're doing fantastic work!

### DIFF
--- a/test/fake_cloud_firestore_test.dart
+++ b/test/fake_cloud_firestore_test.dart
@@ -1367,10 +1367,9 @@ void main() {
       var newMap = {
         'testVal': {'innerKey': 'innerVal'}
       };
-      await firestore
-          .collection('testCollection')
-          .doc('testDoc')
-          .update(newMap);
+      expect(
+          firestore.collection('testCollection').doc('testDoc').update(newMap),
+          completes);
     });
   });
 }
--- a/test/fake_cloud_firestore_test.dart
+++ b/test/fake_cloud_firestore_test.dart
@@ -1354,6 +1354,24 @@ void main() {
         expect(snapshot.docs.first.data()!.title, equals(MovieTitle));
       }));
     });
+
+    test('Replace Null with Map', () async {
+      final firestore = FakeFirebaseFirestore();
+
+      var testMap = {'testVal': null};
+      await firestore
+          .collection('testCollection')
+          .doc('testDoc')
+          .update(testMap);
+
+      var newMap = {
+        'testVal': {'innerKey': 'innerVal'}
+      };
+      await firestore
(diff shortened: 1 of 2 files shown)


test/fake_cloud_firestore_test.dart
```diff
@@ -1354,6 +1354,24 @@ void main() {
...
+      var testMap = {'testVal': null};
+      await firestore
+          .collection('testCollection')
+          .doc('testDoc')
+          .update(testMap);
+
+      var newMap = {
+        'testVal': {'innerKey': 'innerVal'}
+      };
+      await firestore
```
atn832, line 1370: A unit test should have some expectation. In your test, you implicitly expect the update not to throw an exception. But it won't be clear to an uninformed reader. To make it explicit, you could write it as:

```dart
expect(firestore....update(newMap), completes);
```

or possibly (but less readable):
```dart
expect(() async => await firestore...update(newMap), returnsNormally);
```

Note: I haven't tested this myself.

See
* https://api.flutter.dev/flutter/package-test_api_expect/completes.html
* https://api.flutter.dev/flutter/package-matcher_matcher/returnsNormally-constant.html.

atn832 (changes requested):
First off, thanks for choosing this open-source project to contribute for the first time! Your PR title, description and fix look good. I only have one minor tweak for the unit test.

test/fake_cloud_firestore_test.dart
Sapphirewhale: Thanks for the feedback, I wasn't sure what the best practice for an expect statement was, when I'm just expecting the function to complete without an exception, so I appreciate the info. Hopefully this is now fixed

atn832:
I just released your fix to https://pub.dev/packages/fake_cloud_firestore/changelog#123. Thank you again!

Issue #152, opened 2021-04-10
MockDocumentReference update() should throw error when document does not exist


lachiejames:
From the `cloud_firestore` docs:
![image](https://user-images.githubusercontent.com/47808687/114259645-ccbd9f00-9a12-11eb-9ec5-bddd1859d621.png)

The key phrase here being: `If no document exists yet, the update will fail.`

Following this logic, adding this test to `cloud_firestore_mocks_test.dart` should result in a success:
```
  test('updating document should throw error for non-existant IDs', () async {
    final firestore = MockFirestoreInstance();

    expect(
      () async => await firestore.collection('messages').doc('new ID').update({'user.counter': 5}),
      throwsA(isA<NoSuchMethodError>()),
    );
  });
```

But instead, it results in the following test failure:
```
Expected: throws <Instance of 'NoSuchMethodError'>
  Actual: <Closure: () => Future<void>>
   Which: returned a Future that emitted <null>

package:test_api                             expect
main.<fn>
test\cloud_firestore_mocks_test.dart:979
main.<fn>
test\cloud_firestore_mocks_test.dart:976
```

TommyLCox:
Yes, the `update` method in the mock is currently acting like `set`. It creates the document if the document id does not exist. This does not match the docs for the Firestore update method. I assume that Firestore itself is behaving according to its own docs for `set` and `update`.

PR #227, opened 2022-07-01
`DocumentReference.update` throws an exception when the document does not exist

### LINKED ISSUE #40: UpdateData should throw PlatformException if no existing document
As in the #39 's test case, updateData should throw PlatformException if there's no document.

https://github.com/atn832/cloud_firestore_mocks/pull/39/files#diff-bc4857fb99d367ed497ed395a5dfa274R98

```
flutter: 00:01 +7 -2: Firestore behavior comparison: Nested objects update (Cloud Firestore) [E]
flutter:   PlatformException(Error 5, FIRFirestoreErrorDomain, No document to update:
projects/flutter-firestore/databases/(default)/documents/messages/UYK7FkXENdf7mshYHrjR)
```

atn832:
Fixes #40 and #152.

### DIFF
--- a/lib/src/mock_collection_reference.dart
+++ b/lib/src/mock_collection_reference.dart
@@ -159,17 +159,7 @@ class MockCollectionReference<T extends Object?> extends MockQuery<T>
   @override
   Future<DocumentReference<T>> add(T data) async {
     final documentReference = doc();
-    // DocumentReference.update expects a Map<String, Object?>. See
-    // https://pub.dev/documentation/cloud_firestore/2.1.0/cloud_firestore/DocumentReference/update.html.
-    if (data is Map<String, Object?>) {
-      await documentReference.update(data);
-    } else if (_converter != null) {
-      // Use the converter.
-      await documentReference.update(_converter!.toFirestore(data, null));
-    } else {
-      throw StateError('This should never happen');
-    }
-
+    await documentReference.set(data);
     _firestore.saveDocument(documentReference.path);
     QuerySnapshotStreamManager().fireSnapshotUpdate(firestore, path);
     return documentReference;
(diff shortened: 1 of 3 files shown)


lib/src/mock_collection_reference.dart
```diff
@@ -159,17 +159,7 @@ class MockCollectionReference<T extends Object?> extends MockQuery<T>
...
-    if (data is Map<String, Object?>) {
-      await documentReference.update(data);
-    } else if (_converter != null) {
-      // Use the converter.
-      await documentReference.update(_converter!.toFirestore(data, null));
-    } else {
-      throw StateError('This should never happen');
-    }
-
+    await documentReference.set(data);
```
atn832, line 162: `add` sets data for a new document, so we shouldn't use `update`.

test/fake_cloud_firestore_test.dart
```diff
@@ -824,7 +839,7 @@ void main() {
     final foo = firestore.collection('users').doc('foo');
     await foo.set(<String, dynamic>{'name.firstName': 'OldValue Foo'});
     final bar = firestore.collection('users').doc('bar');
-    await foo.set(<String, dynamic>{'name.firstName': 'OldValue Bar'});
+    await bar.set(<String, dynamic>{'name.firstName': 'OldValue Bar'});
```
atn832, line 842: This was a pre-existing bug. I caught it when making `update` throw exceptions on non-existent documents. On line 831 below, bar was actually still unset and `batch.update(bar` threw an exception.

atn832:
Published to https://pub.dev/packages/fake_cloud_firestore/changelog#124.

atn832:
Note: we actually make the Future fail. Will switch to `throw`ing an exception if necessary.

PR #233, opened 2022-08-14
Support DateTime in other Query.where operators

### LINKED ISSUE #228: Support DateTime in other Query.where operators
The way it is implemented now, only range queries support ```DateTime```, though in reality Firestore supports passing ```DateTime``` to most operators, if not all of them.

I suppose you could go about this the same way you implemented range queries, that is by checking if, e.g. ```isEqualTo``` is ```DateTime``` and if so, changing it to ```Timestamp```. Just tell me what you think and I might make a PR with these changes.

Edit:
Forgot to add that the reason range queries work while others don't is because you explicitly change every ```DateTime``` to ```Timestamp``` when setting the document data, hence the need for changing it in queries too.

https://github.com/atn832/fake_cloud_firestore/blob/61ef23cdc5bce6930d394ad770bd61d2fccdef8d/lib/src/mock_document_reference.dart#L150-L152

Zohenn:
**Summary**

- made sure that passing ```DateTime``` works with every applicable ```Query.where``` operator. I tested these types of queries on a real Firestore database and they all should work.
- improved support for data with lists. I don't know if this handles all the cases, but it fixed an issue in one of my projects.
- added tests for queries with ```DateTime```.

Fixes #228

### DIFF
--- a/lib/src/mock_document_reference.dart
+++ b/lib/src/mock_document_reference.dart
@@ -148,21 +148,8 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
       final fieldValue = fieldValuePlatform.value;
       fieldValue.updateDocument(document, key);
     } else {
-      document[key] = _transformValue(value);
+      document[key] = transformDates(value);
     }
-  }
-
-  dynamic _transformValue(dynamic value) {
-    if (value is Map<String, dynamic>) {
-      return value.map((k, v) => MapEntry(k, _transformValue(v)));
-    }
-    if (value is Iterable) {
-      return value.map((e) => _transformValue(e)).toList();
-    }
-    if (value is DateTime) {
-      return Timestamp.fromDate(value);
-    }
-    return value;
   }
 
   Map<String, dynamic> _findNestedDocumentToUpdate(String key) {
--- a/lib/src/mock_document_reference.dart
+++ b/lib/src/mock_document_reference.dart
@@ -147,11 +147,22 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
       final fieldValuePlatform = valueDelegate as MockFieldValuePlatform;
       final fieldValue = fieldValuePlatform.value;
       fieldValue.updateDocument(document, key);
-    } else if (value is DateTime) {
-      document[key] = Timestamp.fromDate(value);
     } else {
-      document[key] = value;
+      document[key] = _transformValue(value);
     }
+  }
+
+  dynamic _transformValue(dynamic value) {
+    if (value is Map<String, dynamic>) {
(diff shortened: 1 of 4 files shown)


lib/src/mock_document_reference.dart
```diff
@@ -147,13 +147,24 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
...
       fieldValue.updateDocument(document, key);
-    } else if (value is DateTime) {
-      document[key] = Timestamp.fromDate(value);
     } else {
-      document[key] = value;
+      document[key] = _transformValue(value);
     }
   }
 
+  dynamic _transformValue(dynamic value) {
```
atn832, line 155: `_transformValue` is quite a vague name, so another developer wouldn't know what it does at first glance. Also, it is the same as `_transformDates` in the other file, except this one supports Maps as well. Let's make it a reusable function instead of duplicating code.

Would you mind moving this function to https://github.com/atn832/fake_cloud_firestore/blob/master/lib/src/util.dart, renaming it to `_transformDates` and document it to explain that it will also recursively convert `DateTime` to `Timestamp` in nested structures such as maps and lists?

Then you can have both mock_document_reference.dart and mock_query.dart use it.

lib/src/mock_query.dart
```diff
@@ -244,6 +244,16 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
     return MockQuery<T>(this, operation);
   }
 
+  dynamic _transformDates(dynamic value) {
+    if(value is DateTime) {
```
atn832, line 248: This is missing a space between `if` and `(`, and wouldn't pass the formatting test workflow. Consider setting up your IDE to format on save to avoid those issues.

atn832 (changes requested):
The code and tests look good. Would you mind turning the 2 methods into a reusable method?

lib/src/mock_document_reference.dart
Zohenn: Good catch, I overlooked this file.

lib/src/mock_query.dart
Zohenn: Well, lately my Android Studio refuses to work with me most of the time, so I can't rely on it too much. Should be ok now.

atn832: Visual Studio Code works well for me, it formats on Save. Alternatively you could run `dart format .` manually.

https://dart.dev/tools/dart-format

atn832:
Thank you! I'll release an update with your code in a few days.

atn832:
Just published your PR at https://pub.dev/packages/fake_cloud_firestore/changelog#131. Thanks again!

PR #235, opened 2022-08-20
Add default Query.limit implementation

### LINKED ISSUE #198: 'FakeConvertedQuery<Memo>' has no instance method 'limit' with matching arguments
In my test, I catched following error.
```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following NoSuchMethodError was thrown building MemoListBody(dependencies:
[UncontrolledProviderScope], state: _ConsumerState#16c8c):
Class 'FakeConvertedQuery<Memo>' has no instance method 'limit' with matching arguments.
Receiver: Instance of 'FakeConvertedQuery<Memo>'
Tried calling: limit(21)
Found: limit(int) => Query<X0>
```

I use [FirestoreQueryBulder](https://firebase.flutter.dev/docs/ui/firestore#advanced-configuration) in widget and following is my code.

```dart
FirestoreQueryBuilder<Memo>(
      pageSize: 20,
      query: ref
          .watch(firebaseFirestoreProvider)
          .collection('users')
          .doc(uid)
          .collection('memos')
          .orderBy('updatedAt', descending: true)
          .withConverter<Memo>(
            fromFirestore: (snapshot, _) => Memo.fromDocumentSnapshot(snapshot),
            toFirestore: (obj, _) => obj.toJson(),
          ),
      builder: (context, snapshot, _) {...});
```

And following is test code.
```dart
 group('when memo list is empty', () {
      testWidgets('show empty text', (tester) async {
        final firestore = FakeFirebaseFirestore();
        final mockUser = MockUser(uid: 'test');
        final mockAuthState = MockAuthState();
        when(mockAuthState.firebaseUser).thenReturn(AsyncData(mockUser));
        final mockAuthController = MockAuthController(mockAuthState);
        await tester.pumpWidget(
          wrapWithMaterial(
            ProviderScope(
              overrides: [
                firebaseFirestoreProvider.overrideWithValue(firestore),
                authControllerProvider.overrideWithValue(mockAuthController),
              ],
              child: const MemoListPage(),
            ),
          ),
        );
        await tester.idle();
        await tester.pump();
        expect(find.text('No data.'), findsOneWidget);
      });
    });
```

I'm wondering if this error occurs when using both WithConverter and limit, but I don't know how to solve it.

Here is my whole code.
https://github.com/oke331/memo_sample/tree/develop

xni06:
`Query.limit` has not been implemented thus resulting in the following:
```
Class 'FakeConvertedQuery<Job>' has no instance method 'limit' with matching arguments.
Receiver: Instance of 'FakeConvertedQuery<Job>'
Tried calling: limit(11)
Found: limit(int) => Query<X0>
```

This PR provides a default `limit` implementation.

Fixes #198

### DIFF
--- a/lib/src/fake_converted_query.dart
+++ b/lib/src/fake_converted_query.dart
@@ -33,5 +33,8 @@ class FakeConvertedQuery<T extends Object?> extends FakeQueryWithParent<T> {
   FakeQueryWithParent? get parentQuery => _nonConvertedParentQuery;
 
   @override
+  Query<T> limit(int limit) => this;
+
+  @override
   dynamic noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
 }
--- a/lib/src/fake_converted_query.dart
+++ b/lib/src/fake_converted_query.dart
@@ -33,7 +33,11 @@ class FakeConvertedQuery<T extends Object?> extends FakeQueryWithParent<T> {
   FakeQueryWithParent? get parentQuery => _nonConvertedParentQuery;
 
   @override
-  Query<T> limit(int limit) => this;
+  Query<T> limit(int limit) {
+    return FakeConvertedQuery<T>(
+        _nonConvertedParentQuery.limit(limit) as FakeQueryWithParent,
+        _converter);
+  }
 
   @override
   dynamic noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -24,7 +24,6 @@ dev_dependencies:
     sdk: flutter
   pedantic: ^1.9.0
   test: ^1.15.2
-  mocktail: ^0.3.0
 
 false_secrets:
   - google-services.json
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -24,6 +24,7 @@ dev_dependencies:
     sdk: flutter
   pedantic: ^1.9.0
   test: ^1.15.2
+  mocktail: ^0.3.0
 
 false_secrets:
   - google-services.json
--- a/test/src/fake_converted_query_test.dart
+++ b/test/src/fake_converted_query_test.dart
@@ -1,0 +1,42 @@
+import 'package:fake_cloud_firestore/src/converter.dart';
+import 'package:fake_cloud_firestore/src/fake_converted_query.dart';
+import 'package:fake_cloud_firestore/src/fake_query_with_parent.dart';
+import 'package:fake_cloud_firestore/src/mock_query.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mocktail/mocktail.dart';
+
+// ignore: subtype_of_sealed_class
+class MockFakeQueryWithParent<T> extends Mock implements MockQuery<T> {}
+
+class MockConverter<T> extends Mock implements Converter<T> {}
+
+void main() {
+  test('FakeQueryWithParentImpl.limit should throw NoSuchMethodError', () {
+    expect(
+      () => FakeQueryWithParentImpl().limit(1),
+      throwsA(isA<NoSuchMethodError>()),
+    );
+  });
+
+  test('FakeConvertedQuery.limit should return normally', () {
+    // arrange
+    final mockFakeQueryWithParent =
+        MockFakeQueryWithParent<Map<String, dynamic>>();
+    final mockConverter = MockConverter<Map<String, dynamic>>();
+
+    // act
+    final query = FakeConvertedQuery<Map<String, dynamic>>(
+      mockFakeQueryWithParent,
+      mockConverter,
+    );
+
+    // assert
+    expect(() => query.limit(1), returnsNormally);
+  });
+}
+
+// ignore: subtype_of_sealed_class
+class FakeQueryWithParentImpl extends FakeQueryWithParent {
+  @override
+  dynamic noSuchMethod(Invocation invocation) => super.noSuchMethod(invocation);
+}
--- a/test/src/fake_converted_query_test.dart
+++ b/test/src/fake_converted_query_test.dart
@@ -11,13 +11,6 @@ class MockFakeQueryWithParent<T> extends Mock implements MockQuery<T> {}
 class MockConverter<T> extends Mock implements Converter<T> {}
 
 void main() {
-  test('FakeQueryWithParentImpl.limit should throw NoSuchMethodError', () {
-    expect(
-      () => FakeQueryWithParentImpl().limit(1),
-      throwsA(isA<NoSuchMethodError>()),
-    );
-  });
-
   test('FakeConvertedQuery.limit should return normally', () {
     // arrange
     final mockFakeQueryWithParent =
--- a/test/src/fake_converted_query_test.dart
+++ b/test/src/fake_converted_query_test.dart
@@ -4,7 +4,7 @@ import 'package:flutter_test/flutter_test.dart';
 import '../fake_cloud_firestore_test.dart';
 
 void main() {
-  test('`limit` clips results', () async {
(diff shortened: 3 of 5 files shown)


atn832:
Thank you for your PR @xni06. Would you mind writing a unit test to verify that it works, and to prevent future users from breaking your use-cases?

lib/src/fake_converted_query.dart
```diff
@@ -32,6 +32,9 @@ class FakeConvertedQuery<T extends Object?> extends FakeQueryWithParent<T> {
   @override
   FakeQueryWithParent? get parentQuery => _nonConvertedParentQuery;
 
+  @override
+  Query<T> limit(int limit) => this;
```
atn832, line 36: After looking more closely to the current implementation, I realize that `MockQuery` already implements `limits` and does so correctly: it actually clips results.
https://github.com/atn832/fake_cloud_firestore/blob/c71a47625b5f819fce8fab9f30b2b8a50e23ff02/lib/src/mock_query.dart#L195

We even have a unit test for it:
https://github.com/atn832/fake_cloud_firestore/blob/c71a47625b5f819fce8fab9f30b2b8a50e23ff02/test/mock_query_test.dart#L173

Since `MockQuery` and `FakeConvertedQuery` both extend `FakeQueryWithParent`, the correct way to fix `limit` for converted queries is to move `MockQuery`'s `limit` implementation to `FakeQueryWithParent`. Then `limit` should work for both queries with and without converters.

Would you mind removing this code and moving `MockQuery`'s `limit` implementation to `FakeQueryWithParent`?

test/src/fake_converted_query_test.dart
```diff
@@ -0,0 +1,42 @@
...
+  });
+
+  test('FakeConvertedQuery.limit should return normally', () {
+    // arrange
+    final mockFakeQueryWithParent =
+        MockFakeQueryWithParent<Map<String, dynamic>>();
+    final mockConverter = MockConverter<Map<String, dynamic>>();
+
+    // act
+    final query = FakeConvertedQuery<Map<String, dynamic>>(
```
atn832, line 28: As a rule, our unit tests do not refer to the internal implementations. The only reference we do to the project is when instantiating `FakeFirebaseFirestore`. This is to ensure that the project correctly fakes Firestore.

You can copy/paste this test instead:
```dart
import 'fake_cloud_firestore_test.dart';



  test('`limit` clips results', () async {
    final from = (snapshot, _) => Movie()..title = snapshot['title'];
    final to = (Movie movie, _) => {'title': movie.title};

    final firestore = FakeFirebaseFirestore();

    final moviesCollection = firestore
        .collection('movies')
        .withConverter(fromFirestore: from, toFirestore: to);
    await moviesCollection.add(Movie()..title = 'A long time ago');
    await moviesCollection.add(Movie()..title = 'Robot from the future');
    final searchResults =
        await moviesCollection.limit(1).get();
    expect(searchResults.size, equals(1));
    final movieFound = searchResults.docs.first.data();
    expect(movieFound.title, equals('A long time ago'));
  });
```

pubspec.yaml
```diff
@@ -24,6 +24,7 @@ dev_dependencies:
     sdk: flutter
   pedantic: ^1.9.0
   test: ^1.15.2
+  mocktail: ^0.3.0
```
atn832, line 27: After applying the remarks I left in fake_converted_query_test.dart, you can safely remove the dependency to mocktail.

atn832 (changes requested):

test/src/fake_converted_query_test.dart
```diff
@@ -0,0 +1,42 @@
...
+// ignore: subtype_of_sealed_class
+class MockFakeQueryWithParent<T> extends Mock implements MockQuery<T> {}
+
+class MockConverter<T> extends Mock implements Converter<T> {}
+
+void main() {
+  test('FakeQueryWithParentImpl.limit should throw NoSuchMethodError', () {
+    expect(
+      () => FakeQueryWithParentImpl().limit(1),
+      throwsA(isA<NoSuchMethodError>()),
```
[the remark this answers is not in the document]
xni06: Removed

test/src/fake_converted_query_test.dart
xni06: Replaced

pubspec.yaml
xni06: Done as part of previous commit

lib/src/fake_converted_query.dart
xni06: Doing so results in the following error - could you please advise?

```
Type: FakeQueryWithParent<T>

The argument type 'FakeQueryWithParent<T>' can't be assigned to the parameter type 'MockQuery<T>?'.
(https://dart.dev/diagnostics/argument_type_not_assignable)

Comment (atn832) [reply lib/src/fake_converted_query.dart:36]: You are right. I had forgotten how queries were implemented and what I suggested earlier about moving the code to `FakeQueryWithParent` does not work. I think we should fall back to implementing a custom implementation in `FakeConvertedQuery` like you first did. I tried this, and it seems to work. Would you mind trying this:
```dart
// in fake_converted_query.dart
  @override
  Query<T> limit(int limit) {
    return FakeConvertedQuery<T>(
        _nonConvertedParentQuery.limit(limit) as FakeQueryWithParent,
        _converter);
  }
```
What this does is, if someone calls `convertedQuery.limit(k)`, it will return a converted query whose parent query is the nonConvertedQuery to which limit is applied. This way, we reuse MockQuery's implementation. The unit test passes. I had to tweak the unit test as I found out what I wrote earlier does not work (see comment in the other thread).

In the long term, we should delete `FakeConvertedQuery` and make `MockQuery` support the converted case, just like I did for `MockCollectionReference`. I just filed #236 to keep track of it. This will enable us to support all the other methods to the converted case (startAt, endAt, etc).

test/src/fake_converted_query_test.dart
atn832: After trying it out myself, I realized that `moviesCollection.limit` does not actually run `FakeConverterQuery.limit`. Instead it runs `MockCollectionReference.limit`. In order to force the test of `FakeConverterQuery.limit`, we have to create a query (not a collection), then convert it with a converter. Would you mind replacing it to this test:

```dart
  test('limit clips results', () async {
    final from = (snapshot, _) => Movie()..title = snapshot['title'];
    final to = (Movie movie, _) => {'title': movie.title};

    final firestore = FakeFirebaseFirestore();

    final moviesCollection = firestore
        .collection('movies')
        .withConverter(fromFirestore: from, toFirestore: to);
    await moviesCollection.add(Movie()..title = 'A long time ago');
    await moviesCollection.add(Movie()..title = 'Robot from the future');
    final rawMoviesCollection = firestore.collection('movies');
    final searchResults = await rawMoviesCollection
        .where('title', isNotEqualTo: 'Galactic')
        .withConverter(fromFirestore: from, toFirestore: to)
        .limit(1)
        .get();
    expect(searchResults.size, equals(1));
    final movieFound = searchResults.docs.first.data();
    expect(movieFound.title, equals('A long time ago'));
  });
```

I've verified on my machine that this works as intended: `rawMoviesCollection.where(...)` creates an actual MockQuery, and `MockQuery.withConverter` creates a `FakeConvertedQuery`.

atn832:
Apologies for the complicated review. Queries are one of the trickier parts of the project. I've left two snippets of code that you should be able to use as is, and we should be good to go. Thanks for being so patient.

atn832:
Thank you for your PR! I'll publish an update in a few days after getting some more code in.

atn832:
Just published your PR to https://pub.dev/packages/fake_cloud_firestore/changelog#131. Thanks again @xni06

Issue #234, opened 2022-08-17
FieldValue.arrayRemove doesn't work when using List of maps


volkankahraman:
There are two lists that we use one is a list of strings and the other one is a list of maps
```dart
String eventId = "1";
Map<String, dynamic> attendance = { "id": "1"};

await fakeFirestore.collection('events').doc(eventId).update({
    'attendanceIdList': FieldValue.arrayRemove([attendanceId]),
    'attendanceList': FieldValue.arrayRemove([attendance]),
});
```
This code is removing the `attendanceId` from `attendanceIdList` but it is not working when we remove `attendance` map from the list. Even though they are same as the collection

atn832:
Fixed by #237 and released at  https://pub.dev/packages/fake_cloud_firestore/changelog#132.

Issue #169, opened 2021-05-16
DocumentSnapshot.[] should throw a StateError if a value wasn't found


atn832:
As per the docs. Resolving #165 will solve it. Both should throw that exception.

PR #237, opened 2022-08-26
Fix inconsistencies with original plugin

### LINKED ISSUE #113: directly nested Array should not be supported
this priority is low.
for now, Firestore doesn't support `directly nested array`.

```.js
# on Firestore, directly nested array is "not" supported.
[ [1, 2] , [3, 4] ] # => this data can not be added.

# on Firestore, supported example
[ { "a": [1, 2] }, { "b": [3, 4] } ] # => this data can be saved.
```

But, both of the two cases is supported on cloud_firestore_mocks.
on cloud_firestore_mocks, `[ [1, 2] , [3, 4] ]` can be saved, against real firestore.

### LINKED ISSUE #165: documentSnapshot.get(composite key) should throw an exception when not found
Currently it returns null.

This is what we get from the actual Firestore:
image

### LINKED ISSUE #194: FieldValue.arrayUnion doesn't work properly with arrays of maps
Version: `fake_cloud_firestore: ^1.1.3`

Adding a map to an array using FieldValue.arrayUnion doesn't work properly.
The following test should complete successfully, but it doesn't:
```
test('add map to array if it does not exist already', () async {
  final firestore = FakeFirebaseFirestore();
  await firestore.collection('users').doc('u1').set({'friends' : [
    {'name': 'Harry'},
    {'name': 'Ron'}
  ]});

  await firestore.collection('user').doc('u1').set({'friends' : FieldValue.arrayUnion([
    {'name': 'Harry'},
    {'name': 'Hermione'}
  ])}, SetOptions(merge: true));

  final rawUser = await firestore.collection('users').doc('u1').get();
  expect(rawUser.data(), {'friends': [
    {'name': 'Harry'},
    {'name': 'Ron'},
    {'name': 'Hermione'}
  ]});
});
```
This is the output:
```
Expected: {'friends': [{'name': 'Harry'}, {'name': 'Ron'}, {'name': 'Hermione'}]}
Actual: {'friends': [{'name': 'Harry'}, {'name': 'Ron'}, {'name': 'Harry'}, {'name': 'Hermione'}]}
```
As the output shows, the maps are added to the array `friends`, but the "union part" didn't work. `{'name': 'Harry'}` was added again, even if an equal map exists already. This would not be the case using the "real" Firestore.

With other data types like `String` it works like the "real" one, so no issues here:
```
test('add map to array if it does not exist already', () async {
  final firestore = FakeFirebaseFirestore();
  await firestore.collection('users').doc('u1').set({'friends' : ['Harry', 'Ron']});

  await firestore.collection('user').doc('u1').set(
    {'friends' : FieldValue.arrayUnion(['Harry', 'Hermione'])},
    SetOptions(merge: true),
  );

  final rawUser = await firestore.collection('users').doc('u1').get();
  expect(rawUser.data(), {'friends': ['Harry', 'Ron', 'Hermione']});
});
```

Zohenn:
**Summary**
This PR aims to make this package more consistent with original by solving some of the submitted issues.
* handle maps correctly when using ```FieldValue.arrayUnion``` or ```FieldValue.arrayRemove```, fixes #194 #234. This now works with dates too, though it was not submitted.
* throw when trying to save data with nested array, fixes #113.
* throw ```StateError``` when trying to access non-existent field with ```MockDocumentSnapshot.get```, fixes #165, #169.

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -115,29 +115,35 @@ dynamic transformDates(dynamic value) {
   return value;
 }
 
-bool iterableEqual(Iterable v1, Iterable v2) {
-  if (v1.length != v2.length) {
-    return false;
-  }
-
-  for (var i = 0; i < v1.length; i++) {
-    if (v1.elementAt(i) != v2.elementAt(i)) {
-      return false;
-    }
-  }
-
-  return true;
-}
-
+/// Compares two values, handles Lists and Maps
+/// by recursively calling itself on each entry.
+/// For Lists:
+///   - length has to be the same.
+///   - for each index, value in the other List at that index must be the same.
+/// For Maps:
+///   - keys must be the same.
+///   - for each key, value in the other Map at that key must be the same.
+/// For other types uses default implementation.
 bool deepEqual(dynamic v1, dynamic v2) {
   if (v1 is Map && v2 is Map) {
     return v1.keys.length == v2.keys.length &&
         v1.keys.every((key) => deepEqual(v1[key], v2[key]));
   }
 
-  if (v1 is Iterable && v2 is Iterable) {
-    return iterableEqual(v1, v2);
+  if (v1 is List && v2 is List) {
+    if (v1.length != v2.length) {
+      return false;
+    }
+
+    for (var i = 0; i < v1.length; i++) {
+      if (!deepEqual(v1[i], v2[i])) {
+        return false;
+      }
+    }
+
+    return true;
   }
 
+  // fallback to default comparison
   return v1 == v2;
 }
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -1,4 +1,5 @@
 import 'package:cloud_firestore/cloud_firestore.dart';
+import 'package:collection/collection.dart';
 
 dynamic getSubpath(Map<String, dynamic> root, String path) {
   return _getSubpath(root, path.split('/'));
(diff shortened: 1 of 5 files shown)


atn832:
Thanks for tackling those issues @Zohenn. Btw, in the future, can you please break down the PRs so that each PR fixes only one issue? This makes reviewing the code much easier and faster. And it will also help avoid bugs, since the PR is easier to read.

I will not ask you to break this PR down this time because it's been 5 months. I will review this PR in the next few days.

lib/src/util.dart
```diff
@@ -110,3 +114,36 @@ dynamic transformDates(dynamic value) {
...
+/// Compares two values, handles Lists and Maps
+/// by recursively calling itself on each entry.
+/// For Lists:
+///   - length has to be the same.
+///   - for each index, value in the other List at that index must be the same.
+/// For Maps:
+///   - keys must be the same.
+///   - for each key, value in the other Map at that key must be the same.
+/// For other types uses default implementation.
+bool deepEqual(dynamic v1, dynamic v2) {
```
atn832, line 127: How is this different from https://api.flutter.dev/flutter/package-collection_collection/DeepCollectionEquality-class.html? Can you use this instead?

Zohenn: Thanks, did not know about this part of ```collection``` package.
I suppose this can (and should) be used, will check on that today.

Zohenn: This seems to work perfectly. I kept the utility method though, for convenience.

atn832: Btw, you could keep it by writing something a tiny bit more concise:
```dart
final deepEqual = DeepCollectionEquality().equals;
```
Up to you.

test/fake_cloud_firestore_test.dart
```diff
@@ -614,6 +615,54 @@ void main() {
...
+            FieldValue.arrayUnion([DateTime(2022, 1, 1), DateTime(2022, 1, 2)]),
+      });
+
+      final snapshot = await docRef.get();
+      final dates = (snapshot.get('dates') as List<dynamic>)
+          .cast<Timestamp>()
+          .map((e) => e.toDate())
+          .toList();
+
+      expect(dates, [DateTime(2022, 1, 1), DateTime(2022, 1, 2)]);
```
atn832, line 663: Thank you for catching this use case.

test/fake_cloud_firestore_test.dart
```diff
@@ -876,6 +1018,30 @@ void main() {
     }
   });
 
+  test('Should throw Error if data contains nested list', () async {
```
atn832, line 1021: Since the two `expectLater`'s do not need to be done in order, it looks like you can safely remove `async` and `await`, and the `test` will wait for everything to complete.

See the examples at https://pub.dev/packages/test#future-matchers.

test/fake_cloud_firestore_test.dart
```diff
@@ -694,6 +796,46 @@ void main() {
...
+      final doc = collection.doc('test');
+      await doc.set({
+        'nested': {'field': null},
+        'field': null
+      });
+
+      final snapshot = await doc.get();
+
+      expect(snapshot.get('field'), isNull);
+      expect(snapshot.get('nested.field'), isNull);
```
atn832, line 826: Nice

atn832 (approved):
Looks great. Thank you @Zohenn for fixing those long-standing issues.

atn832:
Thank you for the high quality PR. I just published your fixes at https://pub.dev/packages/fake_cloud_firestore/changelog#132.

PR #250, opened 2022-11-08
Fix #180 'Null' is not a subtype of type 'Comparable' in type cast issue

### LINKED ISSUE #180: Exception when using orderBy on null field
If you execute a query using orderBy, and the the value of the field that is used to order equals null on some of the documents, then an exception is thrown on mock_query.dart, line 74:
  type 'Null' is not a subtype of type 'Comparable<dynamic>' in type cast

In the real Firestore, there isn't an exception thrown in this case.

ming-chu:
Fix #180

### DIFF
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -83,6 +83,9 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
             try {
               value1 = d1.get(field) as Comparable;
             } catch (error) {
+              // This catch catches the case when the key/value does not exist
+              // and the case when the value is null, and as a result not a
+              // Comparable.
               value1 = null;
             }
           } else if (field == FieldPath.documentId) {
@@ -93,6 +96,8 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
             try {
               value2 = d2.get(field);
             } catch (error) {
+              // This catch catches only the case when the key/value does not
+              // exist.
               value2 = null;
             }
           } else if (field == FieldPath.documentId) {
--- a/lib/src/mock_query.dart
+++ b/lib/src/mock_query.dart
@@ -80,13 +80,21 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
 
           dynamic value1;
           if (field is String) {
-            value1 = d1.get(field) as Comparable;
+            try {
+              value1 = d1.get(field) as Comparable;
+            } catch (error) {
+              value1 = null;
+            }
           } else if (field == FieldPath.documentId) {
             value1 = d1.id;
           }
           dynamic value2;
           if (field is String) {
-            value2 = d2.get(field);
+            try {
+              value2 = d2.get(field);
+            } catch (error) {
+              value2 = null;
+            }
           } else if (field == FieldPath.documentId) {
             value2 = d2.id;
           }
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -312,7 +312,9 @@ void main() {
     ));
   });
 
-  test('orderBy descending returns documents with null fields and missing fields first', () async {
+  test(
+      'orderBy descending returns documents with null fields and missing fields first',
+      () async {
     final instance = FakeFirebaseFirestore();
     await instance.collection('usercourses').add(
         {'name': 'Tom', 'completed_at': Timestamp.fromDate(DateTime.now())});
--- a/test/mock_query_test.dart
+++ b/test/mock_query_test.dart
@@ -296,10 +296,10 @@ void main() {
 
   test('orderBy returns documents with null fields first', () async {
     final instance = FakeFirebaseFirestore();
+    await instance.collection('usercourses').add({'completed_at': null});
     await instance
         .collection('usercourses')
         .add({'completed_at': Timestamp.fromDate(DateTime.now())});
-    await instance.collection('usercourses').add({'completed_at': null});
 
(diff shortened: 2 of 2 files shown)


test/mock_query_test.dart
```diff
@@ -312,6 +312,35 @@ void main() {
     ));
   });
 
+  test('orderBy descending returns documents with null fields first', () async {
```
atn832, line 315: I'd rename this as "with null fields and missing fields first"

lib/src/mock_query.dart
```diff
@@ -80,13 +80,21 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
 
           dynamic value1;
           if (field is String) {
-            value1 = d1.get(field) as Comparable;
+            try {
+              value1 = d1.get(field) as Comparable;
+            } catch (error) {
```
atn832, line 85: This catch catches the case when the key/value does not exist and the case when the value is null, and as a result not a Comparable. A comment would help future readers what is happening.

lib/src/mock_query.dart
```diff
@@ -80,13 +80,21 @@ class MockQuery<T extends Object?> extends FakeQueryWithParent<T> {
...
+            }
           } else if (field == FieldPath.documentId) {
             value1 = d1.id;
           }
           dynamic value2;
           if (field is String) {
-            value2 = d2.get(field);
+            try {
+              value2 = d2.get(field);
+            } catch (error) {
```
atn832, line 95: This catch catches only the case when the key/value does not exist.

test/mock_query_test.dart
atn832, line 315: I have a question on the real Firestore behavior though. Does `descending: true` and `descending: false` both return null values first? Cause the way we implemented `orderBy` in fake_cloud_firestore, it returns null values first no matter the value of `descending`. See https://github.com/atn832/fake_cloud_firestore/blob/01e77ce065d6648a5dd8ba31b8296dd34217a5d2/lib/src/mock_query.dart#L93-L102

atn832:
This bug was much more complicated than I thought. Thank you so much for fixing it.

There were two issues. First #237 recently introduces a change where `doc.get([missing key])` would throw an exception. So ordering would fail if some docs were missing values.

Second, we would convert `value1` to a `Comparable`, while not converting `value2` to a `Comparable`, since `Comparable.compareTo` does not require the second value to be a `Comparable` itself. So our pre-existing tests with null values didn't catch teh bug. It passed only because the first doc in our test happened to have a non-null value, while the second doc had a null value. https://github.com/atn832/fake_cloud_firestore/blob/01e77ce065d6648a5dd8ba31b8296dd34217a5d2/test/mock_query_test.dart#L297

ming-chu:
@atn832 
Thanks for the review, I will update it soon with code comments and a better test name.😃

test/mock_query_test.dart
ming-chu: Let me check 🤔

ming-chu: Looks like the doc with null value on the order field will skip and will not list out😅

ming-chu: FYR: https://firebase.google.com/docs/firestore/query-data/order-limit-data

Note: An orderBy() clause also filters for existence of the given field. The result set will not include documents that do not contain the given field.

atn832: WOW. Good find

ming-chu: I also got this:

https://firebase.google.com/docs/firestore/manage-data/data-types

Value type ordering

When a query involves a field with values of mixed types, Cloud Firestore uses a deterministic ordering based on the internal representations. The following list shows the order:

Null values
Boolean values
Integer and floating-point values, sorted in numerical order
Date values
Text string values
Byte values
Cloud Firestore references
Geographical point values
Array values
Map values

atn832: In order to keep this PR simple and to quickly and safely fix #180, maybe we could make a few separate PRs:
- in this PR, you can fix the exception when a null value is converted to a Comparable. You can then simplify your unit tests to not include docs with missing attributes. Perhaps you could just reuse the existing test at https://github.com/atn832/fake_cloud_firestore/blob/01e77ce065d6648a5dd8ba31b8296dd34217a5d2/test/mock_query_test.dart#L297 and put the doc with null value in first position.
- in a separate PR, we can filter out docs with missing attributes.
- in the far future, we could separately handle whether the order for mixed types is inverted when using descending or not (if `descending: true`, do null values appear first or last?), and handle mixed types more exhaustively.

ming-chu:
@atn832 I resolved the issues you mentioned, 
but the following test seems is not the correct behavior of Firestore that will not return documents with missed ordering field:
`orderBy descending returns documents with null fields and missing fields first`
should I keep this test or remove it?

atn832:
You can remove `orderBy descending returns documents with null fields and missing fields first` altogether to keep it clean. We can cover `descending: true` and the actual behavior for missing fields in a separate PR. Thank you!

ming-chu:
> You can remove `orderBy descending returns documents with null fields and missing fields first` altogether to keep it clean. We can cover `descending: true` and the actual behavior for missing fields in a separate PR. Thank you!

Agree, let's keep it clean. Thank you!

atn832:
Perfect! I'll release a new version in a few days.

atn832:
Took me a bit more than a few days after all. Just published your fix to https://pub.dev/packages/fake_cloud_firestore/changelog#210. Thanks @ming-chu!

PR #251, opened 2022-11-11
Fix FieldValue equality

Zohenn:
Base library makes it possible to compare ```FieldValue.delete()``` and ```FieldValue.serverTimestamp()```. The moment you call ```FakeFirebaseFirestore()``` it stops working, because it overwrites global ```FieldValueFactoryPlatform.instance``` with ```MockFieldValueFactoryPlatform```.

I added a test, it's basically the same as that one:
https://github.com/firebase/flutterfire/blob/cbb24aca7b1cfb264005e069e4377710a0f7b67f/packages/cloud_firestore/cloud_firestore/test/field_value_test.dart

### DIFF
--- a/test/fake_cloud_firestore_test.dart
+++ b/test/fake_cloud_firestore_test.dart
@@ -534,6 +534,16 @@ void main() {
   });
 
   group('FieldValue', () {
+    test('Equality', () {
+      // This is needed, so that the FieldValueFactoryPlatform.instance
+      // is replaced with MockFieldValueFactoryPlatform
+      FakeFirebaseFirestore();
+      expect(FieldValue.delete() == FieldValue.delete(), isTrue);
+      expect(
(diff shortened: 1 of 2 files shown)


test/fake_cloud_firestore_test.dart
```diff
@@ -534,6 +534,16 @@ void main() {
   });
 
   group('FieldValue', () {
+    test('Equality', () {
```
atn832, line 537: Would you mind adding the permalink to the source tests? This way we can easily adapt if they change their tests in the future.

atn832:
Looks good. I'll just merge now. Feel free to add the permalink comment in a separate PR.

test/fake_cloud_firestore_test.dart
Zohenn: Sure, will do.

PR #253, opened 2022-11-26
support metadata changes

danReynolds:
Adds support for specifying metadata changes coming from the cache or not like the cloud_firestore API supports.


atn832:
Thanks for putting in all that work @danReynolds. Before I review the PR, would you mind writing unit tests to test all the new changes you've made? This helps with the review because it shows how it's used, and also ensures that no one accidentally breaks your code in the future.

danReynolds:
Yep! Added some @atn832 thanks. Let me know if there's anything else you need.

lib/src/fake_converted_query.dart
```diff
@@ -26,7 +26,10 @@ class FakeConvertedQuery<T extends Object?> extends FakeQueryWithParent<T> {
                 toFirestore: _converter.toFirestore)
             .get())
         .toList();
-    return MockQuerySnapshot(await Future.wait(convertedSnapshots));
+    return MockQuerySnapshot(
```
atn832, line 29: Very nice of you to remember the case of converted queries as well.

atn832:
The tests really helped me navigate your PR. Thanks!

atn832:
Just published your feature to https://pub.dev/packages/fake_cloud_firestore/changelog#210.

PR #262, opened 2023-01-17
Delete reflects on docsData

ga-bri-el:
https://github.com/atn832/fake_cloud_firestore/issues/221

Let me know if one of those tests aren't required. I was curious if it would maintain the sub collections, that's why I wrote one for that case.


atn832:
Those two tests are useful. I usually ask to not use dump to check the contents, but this bug is specifically about dump, so this time it's actually needed. Thanks!

atn832:
Just published your fix to https://pub.dev/packages/fake_cloud_firestore/changelog#240. Thank you!

patrickhisnibrataas:
@atn832 
>  I usually ask to not use dump to check the contents

Why is that? Is it not supposed to reflect the actual contents of the database? Right now I have a situation where the same code running towards the emulator and the fake having differences in what's stored in the database and it was very confusing to see that using dump().

PR #264, opened 2023-02-07
where with FieldPath.documentID and DocumentReference

agent3bood:
Fix:
`FieldPath.documentID` should be matched against `DocumentReference`.

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -121,6 +121,22 @@ dynamic transformDates(dynamic value) {
   return value;
 }
 
+/// Extract the id from DocumentReference.
+/// Handles complex structures like Map and List
+/// by recursively calling itself on each entry.
+dynamic transformDocumentReference(dynamic value) {
+  if (value is Map<String, dynamic>) {
+    return value.map((k, v) => MapEntry(k, transformDates(v)));
+  }
+  if (value is Iterable) {
+    return value.map((e) => transformDocumentReference(e)).toList();
+  }
+  if (value is DocumentReference) {
+    return value.id;
+  }
+  return value;
+}
+
 /// Convenience method for comparison of collections, e.g. Lists, Maps.
 bool deepEqual(dynamic v1, dynamic v2) {
   return DeepCollectionEquality().equals(v1, v2);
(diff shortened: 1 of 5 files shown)


lib/src/util.dart
```diff
@@ -121,6 +121,22 @@ dynamic transformDates(dynamic value) {
   return value;
 }
 
+/// Extract the id from DocumentReference.
+/// Handles complex structures like Map and List
+/// by recursively calling itself on each entry.
+dynamic transformDocumentReference(dynamic value) {
+  if (value is Map<String, dynamic>) {
+    return value.map((k, v) => MapEntry(k, transformDates(v)));
```
atn832, line 129: I believe you want to call `transformDocumentReference` here.

lib/src/util.dart
```diff
@@ -121,6 +121,22 @@ dynamic transformDates(dynamic value) {
   return value;
 }
 
+/// Extract the id from DocumentReference.
+/// Handles complex structures like Map and List
+/// by recursively calling itself on each entry.
+dynamic transformDocumentReference(dynamic value) {
```
atn832, line 127: Instead of copy/pasting this utility, it seems like we could make a reusable method that recursively transforms individual values. How about this:
```dart
/// Transforms a non-recursive value.
typedef TransformSimpleValue = dynamic Function (dynamic simpleValue);

dynamic transformValue(dynamic value, TransformSimpleValue transformSimpleValue) {
if (value is Map<String, dynamic>) {
    return value.map((k, v) => MapEntry(k, transformValue(v, transformSimpleValue)));
  }
  if (value is Iterable) {
    return value.map((e) => transformValue(e, transformSimpleValue)).toList();
  }
  return transformSimpleValue(value);
}
```

Then for your case, you could call it like this:
```dart
isEqualTo = transformValue(isEqualTo, (value) => value is DocumentReference ? value.id : value);
```

And you can refactor the Date transformation in a similar fashion.

agent3bood: Thanks for the feedback, code refactored to use common structure and utils.

atn832:
Thank you! I just released your code to https://pub.dev/packages/fake_cloud_firestore/changelog#2411.

PR #266, opened 2023-02-15
Support for cloud_firestore 4.4.0

sydneyagcaoili:
### What's in this PR

**Problem**
- With the latest cloud_firestore [changes](https://github.com/firebase/flutterfire/commit/9d91d513fad326f9c928d7d96d03e2c031875903), when using this version, signature related errors appeared.
- With [this](https://github.com/firebase/flutterfire/commit/538090fc49078b8d6c484d8db9049f06d05157dd) change, we need to support FieldPath on docRef.update

**Fix**
- convert relevant signatures to match cloud_firestore 's
- Add support for FieldPath params to docRef.update
- Fix test related issues
- Add test for FieldPath support on docRef.update


sydneyagcaoili:
https://github.com/atn832/fake_cloud_firestore/issues/265 relevant issue

atn832:
This looks very good. And thanks for the quick fix.

lib/src/mock_document_reference.dart
```diff
@@ -103,7 +103,7 @@ class MockDocumentReference<T extends Object?> implements DocumentReference<T> {
         message: 'Some requested document was not found.',
       ));
     }
-    return _setRawData(data);
+    return _setRawData(Map.from(data));
```
atn832, line 106: What is this for?

atn832:
Looks like all unit tests still pass. Awesome!

One more thing you could do in this PR is modify the affected pre-existing unit tests by passing iterables. In the Firestore PR you mentioned, they used this trick in their tests:
```dart
// To force the list to be an iterable
whereIn: ['Ready to Ship', 'Ordered'].map((e) => e),
```

We could make a utility function and use it here and there:
```dart
Iterable<R> toIterable<R>(List<R> list) => list.map((e) => e);
```

The tests in question should be in:
- https://github.com/atn832/fake_cloud_firestore/blob/master/test/mock_query_test.dart for the query change
- https://github.com/atn832/fake_cloud_firestore/blob/master/test/fake_cloud_firestore_test.dart for the docRef.update change.

lib/src/mock_document_reference.dart
sydneyagcaoili: Due to new support to FieldPath on **update**, the Map<String, dynamic> is not gonna cut it.. Seems this needs more work.

sydneyagcaoili:
@atn832 Agreed on supporting iterables.

Seems there needs to be more work on supporting FieldPath on update.  Not confident I'll be able to work on it quickly due to main commitments. But can do in my free time when no one else does.

PR #268, opened 2023-03-24
fix: streamController nullcheck to prevent `ConcurrentModificationError`

simpson-peter:
Introduces a null check on `streamController` in `query_snapshot_stream_manager.dart`, which prevents a `ConcurrentModificationError` occurring. I discovered this error while writing a test which utilized three separate create, update, and delete batch commits on the same collcection.


atn832:
Thanks for the fix! It must have been hard to debug. Would you have that unit test handy? It'd be great to add such a test case in our library to prevent similar errors in the future.

Issue #224, opened 2022-04-16
`docChanges` reports incorrect changes.


ccrusius:
The following test fails and shows other incorrect behavior: `docChanges` is not reporting changes but the new contents of the collection. Nothing is ever deleted, and elements are added multiple times.

The correct sequence of events is "add 1" "add 2" "delete 2", but what is seen is "add 1" "add 1" "add 2" "add 1".

This behavior makes applications that use `docChanges` not testable with the fake.

```dart
import 'package:cloud_firestore/cloud_firestore.dart';
import 'package:fake_cloud_firestore/fake_cloud_firestore.dart';
import 'package:flutter/foundation.dart';
import 'package:flutter_test/flutter_test.dart';

void main() {
  test('Deleting should trigger a delete doc change', () async {
    final Set<String> ids = {};
    final db = FakeFirebaseFirestore();
    final sub = db.collection('docs').snapshots().listen((querySnapshot) {
      for (final change in querySnapshot.docChanges) {
        debugPrint('${change.type} ${change.doc.id}');
        switch (change.type) {
          case DocumentChangeType.removed:
            ids.remove(change.doc.id);
            break;
          default:
            ids.add(change.doc.id);
            break;
        }
      }
    });
    addTearDown(sub.cancel);

    await db.collection('docs').doc('1').set(<String, dynamic>{});
    await pumpEventQueue();
    expect(ids, {'1'});

    await db.collection('docs').doc('2').set(<String, dynamic>{});
    await pumpEventQueue();
    expect(ids, {'1', '2'});

    await db.collection('docs').doc('2').delete();
    await pumpEventQueue();
    expect(ids, {'1'});
  });
}
```

ryanhanks-bestow:
I've run into this as well. Has anyone looked into how difficult this would be to fix?

msxenon:
the same

atn832:
Thanks to @KholmatovS for fixing this. It's now published to https://pub.dev/packages/fake_cloud_firestore/changelog#245.

PR #284, opened 2023-12-15
Fix `docChanges` removed, modified and added with proper index values

KholmatovS:
Fix for #224 . 
Also fixed proper snapshot query emitting when document modified/removed preserving old and new indexes.

### DIFF
--- a/lib/src/mock_collection_reference.dart
+++ b/lib/src/mock_collection_reference.dart
@@ -2,10 +2,12 @@ import 'dart:async';
 import 'dart:math';
 
 import 'package:cloud_firestore/cloud_firestore.dart';
+import 'package:collection/collection.dart';
 import 'package:fake_cloud_firestore/fake_cloud_firestore.dart';
 import 'package:fake_cloud_firestore/src/fake_query_with_parent.dart';
 
 import 'converter.dart';
+import 'mock_document_change.dart';
 import 'mock_document_reference.dart';
 import 'mock_query.dart';
 import 'mock_query_snapshot.dart';
@@ -74,12 +76,15 @@ class MockCollectionReference<T extends Object?> extends MockQuery<T>
         return documentReference.get();
       }).toList();
     }
+    final snapshots = await Future.wait(futureDocs);
     return MockQuerySnapshot<T>(
-      (await Future.wait(futureDocs))
-          .where((snapshot) =>
-              _firestore.hasSavedDocument(snapshot.reference.path))
-          .toList(),
+      snapshots.where((element) {
+        return _firestore.hasSavedDocument(element.reference.path);
+      }).toList(),
       options?.source == Source.cache,
+      documentChanges: snapshots.mapIndexed((index, e) {
+        return MockDocumentChange<T>(e, DocumentChangeType.added, oldIndex: -1, newIndex: index);
+      }).toList(),
     );
   }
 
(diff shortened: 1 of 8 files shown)


test/mock_query_test.dart
```diff
@@ -1678,4 +1678,100 @@ void main() {
...
+
+    final expectedEmitOrder = [
+      // beginning
+      [],
+      // after adding doc 1
+      ['Norman'],
+      // after adding doc 2 (ordered by name)
+      ['Gwen', 'Norman'],
+      // after updating doc 2 (Gwen -> Peter, order by name )
+      ['Norman', 'Peter'],
```
atn832, line 1723: Very nice and thorough test case, and great documentation

KholmatovS:
@atn832 resolved merge conflicts

lib/src/mock_collection_reference.dart
```diff
@@ -74,13 +74,15 @@ class MockCollectionReference<T extends Object?> extends MockQuery<T>
...
     }
-    return MockQuerySnapshot<T>(
+    final snapshot = MockQuerySnapshot<T>(
       (await Future.wait(futureDocs))
           .where((snapshot) =>
               _firestore.hasSavedDocument(snapshot.reference.path))
           .toList(),
       options?.source == Source.cache,
     );
+    QuerySnapshotStreamManager().setCacheQuerySnapshot(this, snapshot);
```
atn832, line 84: Seems like you're already updating cacheQuerySnapshot in fireSnapshotUpdate. Why do we need this for each `MockQuery.get`?

lib/src/query_snapshot_stream_manager.dart
```diff
@@ -81,25 +87,118 @@ class QuerySnapshotStreamManager {
...
+    required List<QueryDocumentSnapshot<T>> docsPrior,
+    required List<QueryDocumentSnapshot<T>> docsCurrent,
+  }) {
+    final _docPriorIndex = docsPrior.indexWhere((element) {
+      return element.id == id;
+    });
+    QueryDocumentSnapshot<T>? _docPrior;
+    if (_docPriorIndex != -1) {
+      _docPrior = docsPrior[_docPriorIndex];
+    }
```
atn832, line 163: Since you only need this on line 188, how about inlining it there like you did on line 180 for docsCurrent?

```dart
   else if (_docCurrentIndex == -1 && _docPriorIndex != -1) {
      /// Document is removed.
      return MockDocumentChange<T>(
        docsPrior[_docPriorIndex]!,
```

lib/src/mock_collection_reference.dart
KholmatovS: Yeah actually we can remove query set from `fireSnapshotUpdate`, and only keep in `MockQuery.get`. Because `fireSnapshotUpdate` is only being called when data is being updated/added/deleted, so it means we are not storing initial cache until something is updated, which will result in two consecutive `added` docchange event, first initial from query where cache wasn't set, and second when data is updated and initial cache set but emits added event

atn832 (approved):
Looks great!!

atn832:
Just published your pull requests to https://pub.dev/packages/fake_cloud_firestore/changelog#245. Thanks for the great update!

diegogarciar:
hey, after upgrading to this version my snapshot listener is not receiving events when calling `reference.update()`. I think the <T> might be extra

image

diegogarciar:
Nevermind, I think the issue is somewhere else. I think now we're not checking the type when creating new instances but here we are, so if we use a converter in the app but a simple reference in the tests, I think this will drop the update

PR #285, opened 2023-12-20
Implement `clearPersistence` to wipe fake database

KholmatovS:
Changes:

- Added `clearPersistence()` implementation to clear fake cloud firestore


atn832:
Thanks again!

PR #286, opened 2023-12-20
fix: increased limit `whereIn` and `arrayContainsAny` disjunctions

KholmatovS:
Increased limit on `whereIn` and `arrayContainsAny` queries per updated Firestore docs: 

https://firebase.google.com/docs/firestore/query-data/queries#limitations_2


atn832:
Thank you @KholmatovS !

PR #287, opened 2023-12-27
Feature/support mocking server timestamp

ming-chu:
This PR is going to provide the interface for mocking the server timestamp.
Btw, I am pretty sure that there are lots of better solutions than mine.

Any advice or comments are welcome.

Thank you for maintaining such an amazing plugin. I appreciated it! ❤️

@atn832


lib/src/fake_cloud_firestore_instance.dart
```diff
@@ -36,9 +37,13 @@ class FakeFirebaseFirestore implements FirebaseFirestore {
...
 
-  FakeFirebaseFirestore(
-      {Stream<Map<String, dynamic>?>? authObject, String? securityRules})
-      : securityRules =
+  final FakeServerTimeProvider? fakeServerTimeProvider;
+
+  FakeFirebaseFirestore({
+    Stream<Map<String, dynamic>?>? authObject,
+    String? securityRules,
+    this.fakeServerTimeProvider,
```
atn832, line 45: Instead of passing your own FakeServerTimeProvider, we probably want to use the clock package like the ticket reporter suggested. Clock allows us to later use FakeAsync to pass time artificially as explained at https://pub.dev/packages/clock.

Would you mind reworking the PR by removing FakeServerTimeProvider, adding clock as a dependency, then modifying the constructor like so:

```dart
  final Clock _clock;

  FakeFirebaseFirestore({
    Stream<Map<String, dynamic>?>? authObject,
    String? securityRules,
    Clock? clock,
  }) : securityRules = ..., _clock = clock ?? Clock()
```

ming-chu: Sure, no problem. I will do that😃 @atn832

lib/src/mock_field_value_platform.dart
```diff
@@ -1,22 +1,31 @@
...
+  void setTimeProvider(FakeServerTimeProvider? fakeServerTimeProvider) {
+    _fakeServerTimeProvider = fakeServerTimeProvider;
+  }
+
+  FieldValueServerTimestamp();
 
   @override
   void updateDocument(Map<String, dynamic> document, String key) {
-    document[key] = Timestamp.now();
+    document[key] = now;
```
atn832, line 28: With the Clock always set up, you can rewrite this as

```dart
document[key] = clock.now()
```

https://pub.dev/documentation/clock/latest/clock/Clock/now.html

lib/src/mock_field_value_platform.dart
```diff
@@ -1,22 +1,31 @@
...
   void updateDocument(Map<String, dynamic> document, String key);
 }
 
 class FieldValueServerTimestamp extends FakeFieldValue {
-  const FieldValueServerTimestamp();
+  FakeServerTimeProvider? _fakeServerTimeProvider;
+
+  Timestamp get now => _fakeServerTimeProvider?.now ?? Timestamp.now();
+
+  void setTimeProvider(FakeServerTimeProvider? fakeServerTimeProvider) {
```
atn832, line 20: How about using a getter setter?

```dart
set clock(Clock clock) {
  this._clock = clock;
}
```

test/fake_cloud_firestore_test.dart
```diff
@@ -683,6 +684,44 @@ void main() {
...
+        'created': FieldValue.serverTimestamp(),
+      });
+      final users = await firestore.collection('users').get();
+      final bob = users.docs.first;
+      expect(bob.get('created'), isNotNull);
+      final bobCreated = bob.get('created') as Timestamp;
+      expect(bobCreated, fixedTimestamp);
+    });
+
+    test('FieldValue.serverTimestamp() sets the time with relative time',
```
atn832, line 704: If we use Clock, the first test is probably sufficient and you can remove the second test. If you want to implement a second test, perhaps you could use [fake_async](https://pub.dartlang.org/packages/fake_async) and demonstrate the passing of time like they do in https://pub.dev/packages/clock.

lib/src/fake_server_time_provider.dart
atn832, line 1: You can safely delete this

test/fake_cloud_firestore_test.dart
```diff
@@ -1712,3 +1751,17 @@ void main() {
...
+  CustomServerTimeProvider({required this.hoursAgo});
+
+  @override
+  Timestamp get now {
+    // always return the time that n hours ago from now
+    return Timestamp.fromDate(
+      DateTime.now().subtract(Duration(hours: hoursAgo)),
+    );
+  }
+}
```
atn832, line 1767: You can safely delete this.

atn832 (changes requested):
Thanks for the PR! Would you mind replacing FakeServerTimeProvider with Clock from https://pub.dev/packages/clock?

lib/src/mock_field_value_platform.dart
ming-chu: let me think, I was thinking about how the `FieldValueServerTimestamp` can get the clock from FakeFirebaseFirestore

ming-chu:
@atn832  Thank you for the review and I resolved them according to your suggestions.
Please feel free to check and comment.

Thank you very much!!

test/fake_cloud_firestore_test.dart
ming-chu: I think that will be like:
```dart
final fakeClock = Clock(() => DateTime.now().subtract(Duration(hours: 3)));
```

So, maybe I don't prefer to have this test just for demonstration and probably mess the tests up.
What do you think?🤔

atn832 (approved):
Thanks for the quick update @ming-chu !

atn832:
I made tiny tweaks at https://github.com/atn832/fake_cloud_firestore/commit/1036aa2f85c321d0dbeea0b7f525661bd9c99c46. Basically in this project's classes, we first put private variables, then constructors, then methods and getter/setters.

ming-chu:
> I made tiny tweaks at [1036aa2](https://github.com/atn832/fake_cloud_firestore/commit/1036aa2f85c321d0dbeea0b7f525661bd9c99c46). Basically in this project's classes, we first put private variables, then constructors, then methods and getter/setters.

No problem, I will keep it in mind and follow it next time.
Thank you very much!

PR #291, opened 2024-01-10
Feature/fix aggregate issue with converted queries

ming-chu:
This PR has fixed the aggregate query problem on converted queries.
https://github.com/atn832/fake_cloud_firestore/pull/290#discussion_r1445569625
The error log would be like:
```log
type 'Movie' is not a subtype of type 'Map<String, dynamic>' in type cast
package:fake_cloud_firestore/src/fake_aggregate_query.dart 34:56   FakeAggregateQuery._getAggregateQuerySnapshotPlatform.<fn>
dart:_internal                                                     ListIterator.moveNext
package:fake_cloud_firestore/src/fake_aggregate_query.dart 107:27  FakeAggregateQuery.buildAggregateQueryResponseList
package:fake_cloud_firestore/src/fake_aggregate_query.dart 37:12   FakeAggregateQuery._getAggregateQuerySnapshotPlatform
package:fake_cloud_firestore/src/fake_aggregate_query.dart 21:22   FakeAggregateQuery.get
```

### DIFF
--- a/lib/src/aggregate_type_extension.dart
+++ b/lib/src/aggregate_type_extension.dart
@@ -1,0 +1,16 @@
+import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_interface.dart';
+
+extension AggregateTypeExtension on AggregateType {
+  Type get aggregateFieldType {
+    switch (this) {
+      case AggregateType.sum:
+        return sum;
+      case AggregateType.average:
+        return average;
+      case AggregateType.count:
+        return count;
+      default:
+        throw UnimplementedError('Unknown AggregateType: $this');
+    }
+  }
+}
--- a/lib/src/fake_aggregate_query.dart
+++ b/lib/src/fake_aggregate_query.dart
@@ -5,8 +5,6 @@ import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_inte
     as platform_interface;
 import 'package:collection/collection.dart';
 import 'package:fake_cloud_firestore/src/aggregate_type_extension.dart';
-import 'package:fake_cloud_firestore/src/mock_document_snapshot.dart';
-import 'package:fake_cloud_firestore/src/mock_query_document_snapshot.dart';
 import 'package:flutter/foundation.dart';
 
 import 'fake_aggregate_query_snapshot.dart';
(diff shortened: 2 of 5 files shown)


lib/src/aggregate_type_extension.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:cloud_firestore_platform_interface/cloud_firestore_platform_interface.dart';
+
+extension AggregateTypeExtension on AggregateType {
```
atn832, line 3: Fancy! 👍

lib/src/fake_aggregate_query.dart
```diff
@@ -28,10 +31,18 @@ class FakeAggregateQuery implements AggregateQuery {
   @override
   AggregateQuery count() => _query.count();
 
+  Map<String, dynamic> _getRawDocDataMap(QueryDocumentSnapshot<Object?> s) {
+    if (s is MockQueryDocumentSnapshot && s.snapshot is MockDocumentSnapshot) {
```
atn832, line 35: Why not use an assert statement like you did in the other PR? `flutter test` does fail if the assert fails. If something does fails for unexpected reasons, let it fail so we can fix it properly next time. :)

lib/src/fake_aggregate_query.dart
```diff
@@ -28,10 +31,18 @@ class FakeAggregateQuery implements AggregateQuery {
...
+    }
+    // this line should never execute but leave it here to play safe.
+    return s.data() as Map<String, dynamic>;
+  }
+
   AggregateQuerySnapshotPlatform _getAggregateQuerySnapshotPlatform({
     required QuerySnapshot<Object?> snapshot,
   }) {
-    final dataMaps = snapshot.docs.map((e) => e.data() as Map<String, dynamic>);
+    final dataMaps = snapshot.docs.map((s) => _getRawDocDataMap(s));
```
atn832, line 45: Have you considered using `DocumentSnapshot.get`? Then you don't need to expose the internal map. It's better to use the public API whenever possible.

You would replace
```dart
final value = dataMap[field.field];
```
by
```dart
final value = documentSnapshot.get(field.field);
```

It also has the added benefit to make the loop down there clearer. Instead of this:
```dart
    for (final dataMap in dataMaps) {
      for (final field in fields) {
```
you could write:
```dart
    for (final field in fields) {
      for (final documentSnapshot in documentSnapshots) {
```

Also would you mind renaming `valueMap` into something more meaningful like `aggregateValues`? Since you already use the variable name `value` when referring to a document value, `value` and `valueMap` was a bit confusing.

lib/src/fake_aggregate_query.dart
ming-chu: Oh, it's because I want to check the type with the `is` operator so that we can use the instance method without casting it again.
Let me think and revise it. ;)

lib/src/fake_aggregate_query.dart
ming-chu: Good idea!

ming-chu:
@atn832  Thanks for the review.
I just pushed the new commit according to the comments.
Please let me know if I missed any :)

lib/src/fake_aggregate_query.dart
ming-chu: after using `DocumentSnapshot.get` we don't need this method anymore, haha. Thank you!

atn832 (approved):
Thank you for the PR @ming-chu !

atn832:
I published your fix to https://pub.dev/packages/fake_cloud_firestore/changelog#247.

I also switched to Dart 3 and used pattern matching to rewrite `getAggregateFieldName`: https://github.com/atn832/fake_cloud_firestore/commit/f277193d98665292dc929728dde45a465aeaeefd

https://dart.dev/language/patterns#switch-statements-and-expressions

PR #295, opened 2024-02-16
fix: compatibility for listeners with different converters

### LINKED ISSUE #294: listeners not updating with #284 fix
My app has a FirestoreClient that forces you to use `converters`, the converter transforms the response into an 
`Either<FirestoreFailure,Data>` and it works great, for our testing we simply instantiate `FakeFirebaseFirestore` and write to the collection directly without the use of converters because we don't need to provide an object so its easier for us to tests the scenarios  when data might be missing.

With the introduction of #284, we can no longer test this because the `fireSnapshotUpdate` is checking for the converter type. so if you have a converter set but want to write directly without a converter, the listeners are not triggered. The issue comes because that change introduced the creation of a new copy of the querySnapshot that requires the code to know the type of <T>.

I'm not entirely sure what's the purpose of this, but I would like to propose to add the previous behavior if the <T> condition is not met for now so that we keep backwards compatibility. 

image

https://github.com/atn832/fake_cloud_firestore/blob/dd3586df91af12b74f1ce7516e019baf33b853b4/lib/src/query_snapshot_stream_manager.dart#L102-L137

diegogarciar:
## Why?
Snapshot listeners built with a converter will not hear updates made from a reference without a converter
Fixes #294 

## What?

Re added the previous code we had before #284 when the query type doesn't match, I couldn't think of a better solution but I'm open to ideas.

### DIFF
--- a/test/query_snapshot_stream_manager_test.dart
+++ b/test/query_snapshot_stream_manager_test.dart
@@ -60,7 +60,7 @@ void main() {
       collectionRef = instance.collection('users');
     });
 
-    void validate() {
+    void validateSnapshotsWithoutConverter() {
       expect(
         collectionRef.snapshots(),
         emitsInOrder(
@@ -83,13 +83,10 @@ void main() {
         'name': 'Bob',
       });
 
-      validate();
-
-      await pumpEventQueue();
+      validateSnapshotsWithoutConverter();
       await collectionRef.doc('1').set({
         'name': 'Marie',
       });
-      await pumpEventQueue();
     });
     test('can receive updates starting with converter', () async {
       final converterRef = collectionRef.withConverter<User>(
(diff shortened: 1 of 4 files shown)


atn832:
Hi @diegog-sf thanks for reporting the issue. It's a one line change but every fix needs a test to prevent further regressions. Are you able to write a simple test that reproduces the issue you mentioned? I'll try to find time to look at it more closely this week.

diegog-sf:
Thanks, will do over the weekend

diegogarciar:
hey @atn832 longest weekend ever, but I was finally able to add the tests :)

atn832:
Amazing thank you!! I'll take a look in the next few days. Btw the Analyzer issue doesn't seem related to this PR. I'll check separately later.

test/query_snapshot_stream_manager_test.dart
```diff
@@ -2,37 +2,125 @@ import 'dart:async';
...
+    }
+
+    test('can receive updates', () async {
+      await collectionRef.doc('1').set({
+        'name': 'Bob',
+      });
+
+      validate();
+
+      await pumpEventQueue();
```
atn832, line 88: I tried removing all the `await pumpEventQueue()` and the tests still passed. Why are they needed?

test/query_snapshot_stream_manager_test.dart
```diff
@@ -2,37 +2,125 @@ import 'dart:async';
...
+  group('listener updates', () {
+    late CollectionReference collectionRef;
+    setUp(() {
+      collectionRef = instance.collection('users');
     });
-    // Fire a snapshot update.
-    instance.collection('users').add({
-      'name': 'Steve',
+
+    void validate() {
```
atn832, line 63: Would you mind renaming this into something more meaningful? Maybe `validateSnapshotsWithoutConverter()`.

atn832 (approved):

test/query_snapshot_stream_manager_test.dart
diegog-sf: they're use of habit, I can remove them

diegogarciar:
Feedback addressed, I think I also fixed the analyzer issues

atn832:
Thank you for the bug fix, the nice tests and the analyzer fix! I just published your code to https://pub.dev/packages/fake_cloud_firestore/changelog#252.

Issue #297, opened 2024-03-15
throws null value with composite key query on null data document


KholmatovS:
Making query with composite keys when a doc is created but data hasn't been set yet, it throws null value exception

```
Null check operator used on a null value
```

PR: #298

atn832:
I ended up forgetting to release an update. It's now out: https://pub.dev/packages/fake_cloud_firestore/changelog#249. Thank you!

Issue #156, opened 2021-04-18
MockDocumentReference.set({}), then get() returns null


atn832:
This test fails.

```dart
await instance.collection('users').doc('3').set({});
final doc = await instance.collection('users').doc('3').get();
expect(doc.data(), isNotNull);
```

I'll have to try it on the actual Firestore to see how it should behave.

atn832:
If this is indeed a bug, we should revert the changes in https://github.com/atn832/cloud_firestore_mocks/pull/153/files#diff-9d69ed340c3ab50d969a5011a19f493ade18a72981f8bef6e8a2a878e3820d9e back to `set({})`.

jibbers42:
I just starting using this package and ran into the same issue. I'm curious if you ever verified Firestore behavior - I haven't, but it does feel like this should work.

atn832:
https://pub.dev/packages/fake_cloud_firestore/changelog#249

PR #298, opened 2024-03-15
fix: composite key query on null document

KholmatovS:
Fixes null check operation on null value when queried with composite keys on null data document.

Fix for: #297


atn832:
The analyzer issues are from Flutter 3.19's new rules. We can address them separately

atn832:
Thank you for the fix! It seems like a side effect of #156. I'll see if I can find a more general fix.

PR #299, opened 2024-03-16
fix: set({}) followed by get() returns {}

### LINKED ISSUE #156: MockDocumentReference.set({}), then get() returns null

atn832:
Fixes #156


atn832:
@KholmatovS I kept your unit test and replaced your fix by this one. It covers your use case and also fixes #156.

KholmatovS:
@atn832 great, thank you!

PR #302, opened 2024-04-09
Fixed exception when dump() dumps Document Reference

BuyMyBeard:
When database document contains document reference, Exception will occur [here](https://github.com/atn832/fake_cloud_firestore/blob/6164f7f0e9b28ee41853cd6859d8faba4a7f130f/lib/src/fake_cloud_firestore_instance.dart#L149)

Error thrown is this: 
`
JsonUnsupportedObjectError (Converting object did not return an encodable object: Instance of 'MockDocumentReference<Map<String, dynamic>>')
`
I fixed it by supporting DocumentReference type in myEncode util function.

### DIFF
--- a/lib/src/util.dart
+++ b/lib/src/util.dart
@@ -51,6 +51,8 @@ dynamic myEncode(dynamic item) {
       'latitude': item.latitude,
       'longitude': item.longitude,
     };
+  } else if (item is DocumentReference) {
+    return item.path;
   }
   return item;
 }


atn832:
Thanks for the PR @BuyMyBeard. Would you mind also writing a test to prevent regressions in the future? The tests for `dump()` are located here: https://github.com/atn832/fake_cloud_firestore/blob/6164f7f0e9b28ee41853cd6859d8faba4a7f130f/test/fake_cloud_firestore_test.dart#L28-L208

atn832:
> I'd like to produce some tests, but I can't compile the project after cloning it for some reason. These are the exceptions thrown: image
> 
> Is it because of my Flutter SDK version?

This is because of this PR: https://github.com/atn832/fake_cloud_firestore/pull/306. You probably want to get the latest packages by running `flutter pub get`.

PR #310, opened 2024-06-17
fix: batch commit doesn't properly await for individual commands.

nomtats:


atn832:
Thank you @nomtats ! I just published your fix to https://pub.dev/packages/fake_cloud_firestore/changelog#301.

PR #312, opened 2024-07-25
fix: multiple docs change for parallel saved doc snapshots

KholmatovS:
Changes:
- Fixed the wrong docs change state (missing change info) on the snapshots when docs are saved in parallel

### DIFF
--- a/lib/src/query_snapshot_stream_manager.dart
+++ b/lib/src/query_snapshot_stream_manager.dart
@@ -1,6 +1,7 @@
 import 'dart:async';
 
 import 'package:cloud_firestore/cloud_firestore.dart';
+import 'package:collection/collection.dart';
 import 'package:rxdart/rxdart.dart';
 
 import 'fake_query_with_parent.dart';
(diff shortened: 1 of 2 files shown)


atn832 (approved):

atn832:
Thank you!

lib/src/query_snapshot_stream_manager.dart
```diff
@@ -131,6 +132,22 @@ class QuerySnapshotStreamManager {
         if (docChange != null) {
           documentsChange.add(docChange);
         }
+
+        /// When multiple new doc snapshots added for the current state, we need to fire multiple document changes
+        final newDocsChanges = docsCurrent.where((doc) {
```
atn832, line 137: You're checking changes for documents that have appeared since. So if some document is in `docsPrior` but not `docsCurrent`, this logic won't catch the change:
```typescript
    // Remove in parallel.
    final op3 = db.collection('docs').doc('1').set(<String, dynamic>{
      'age': 36,
    }, SetOptions(merge: true));
    final op4 = db.collection('docs').doc('2').set(<String, dynamic>{
      'age': 40,
    }, SetOptions(merge: true));
    await Future.wait([op3, op4]);
```
Only emits `[DocumentChangeType.removed]`, not `[DocumentChangeType.removed, DocumentChangeType.removed]`.

I'll try and submit a PR for it.

atn832:
Just published your fix and #314 to https://pub.dev/packages/fake_cloud_firestore/changelog#302. Thank you!

PR #314, opened 2024-07-27
support firing multiple removals when done in parallel

atn832:
#312 supported the case when two documents are added in parallel and we want to fire an event that includes both additions. This PR supports any parallel modification, such as when two documents are removed or modified in parallel.

### DIFF
--- a/lib/src/query_snapshot_stream_manager.dart
+++ b/lib/src/query_snapshot_stream_manager.dart
@@ -6,6 +6,7 @@ import 'package:rxdart/rxdart.dart';
 
 import 'fake_query_with_parent.dart';
 import 'mock_document_change.dart';
+import 'mock_query_document_snapshot.dart';
 import 'mock_query_snapshot.dart';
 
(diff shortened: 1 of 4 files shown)


lib/src/query_snapshot_stream_manager.dart
```diff
@@ -120,33 +121,22 @@ class QuerySnapshotStreamManager {
...
-            docsPrior: docsPrior,
-            docsCurrent: docsCurrent,
-          );
-        });
-        if (newDocsChanges.isNotEmpty) {
-          documentsChange.addAll(newDocsChanges.whereNotNull());
-        }
+        // Collect change from the whole query documents, whether they be added,
+        // deleted or modified.
+        final affectedIds = <String>{
```
atn832, line 126: @KholmatovS

KholmatovS: Amazing, thank you!

